PR #20, opened 2021-10-28
Infra svc

maximelb:
## Description of the change

By default Configs now uses the `infrastructure-service` in the cloud to perform the update as it is more reliable and standardized. Legacy behavior is still available using a flag.

Also added support for service requests that impersonate.

Added a small comment on gevent use in the sdk.

Increasing the buffer size downloading Artifacts to speedup transfer.

## Type of change


mireaulf (commented):
lgtm

PR #39, opened 2022-07-08
Fix Spout with Threading and add CLI.

maximelb:
## Description of the change

Fixing issues with Spout since we moved away from gevent. Also made Spouts available via the CLI.

## Type of change


josh-lc (commented):
LGTM

PR #60, opened 2022-10-10
Use service-only for config pushes by default.

maximelb:
## Description of the change

Move the config syncing to use one or the other only, not local and cloud service at the same time. It only muddies the water when an error occurs.

## Type of change


jprogramtech (commented):
LGTM

PR #71, opened 2022-11-16
Allow directory to glob recursively

shortstack:
## Description of the change

`glob.iglob( include )` recursive defaults to `False`, added `glob.iglob( include, recursive=True )` so `**` calls recursively. 

Example, for pushing rules in nested directories:
```
version: 3
include:
  - rules/**/*.yml
```

## Type of change


maximelb:
Oh thanks, good catch!

maximelb (approved):

PR #139, opened 2024-08-19
add arl support to perform set operation against Hive API

ezereynalc:
## Description of the change

Add support to the Hive client for the ARL field

## Type of change

## Related issues

> Fix [#2587](https://github.com/refractionPOINT/tracking/issues/2587)

### DIFF
--- a/limacharlie/Hive.py
+++ b/limacharlie/Hive.py
@@ -71,7 +71,7 @@ class Hive( object ):
         if len( usrMtd ) != 0:
             req[ 'usr_mtd' ] = json.dumps( usrMtd )
         if record.arl is not None:
-            req [ 'arl' ] = json.dumps(record.arl)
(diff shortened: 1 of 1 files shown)


limacharlie/Hive.py
```diff
@@ -70,6 +70,8 @@ def set( self, record ):
             req[ 'etag' ] = record.etag
         if len( usrMtd ) != 0:
             req[ 'usr_mtd' ] = json.dumps( usrMtd )
+        if record.arl is not None:
+            req [ 'arl' ] = json.dumps(record.arl)
```
maximelb, line 74: This should always be a string right?
Not a big deal since json.dumps will produce a string from a string IIRC.

maximelb (approved):

maximelb (approved):

PR #155, opened 2025-02-19
feat: [refractionPOINT/tracking#3025] Add new "limacharlie users invite" action for inviting a user

tomaz-lc:
## Description

This pull request adds new `limacharlie users invite --email=<email>` CLI action for inviting a user.

## Example Usage

```bash
# Invalid usage - 1
$ limacharlie users
usage: limacharlie user [-h] {invite} ...
limacharlie user: error: the following arguments are required: action

# Invalid usage - 2
$ limacharlie users invite
Error: Please provide either --email or --file option.

# Invalid usage - 3
$ limacharlie users invite --email=tomaz@tomaz.me --file=1.txt
Error: --email and --file are mutually exclusive, please provide only one.

# Success - already invited
$ limacharlie users invite --email=tomaz@tomaz.me
User with email tomaz@tomaz.me already exists / has already been invited.

# Success - invite sent
$ limacharlie users invite --email=tomaz+test1@tomaz.me
User with email tomaz+test1@tomaz.me has been invited.

# Invite multiple users using comma delimited --email flag
$ limacharlie users invite --email=tomaz+test1@tomaz.me,tomaz+test2@tomaz.me,tomaz+test3@tomaz.me
Inviting user with email tomaz+test1@tomaz.me.
User with email tomaz+test1@tomaz.me already exists / has already been invited.
Inviting user with email tomaz+test2@tomaz.me.
User with email tomaz+test2@tomaz.me already exists / has already been invited.
Inviting user with email tomaz+test3@tomaz.me.
User with email tomaz+test3@tomaz.me already exists / has already been invited.

# Invite multiple users from new line delimited text file
$ cat 1.txt 
tomaz+test1@tomaz.me
tomaz+test2@tomaz.me
tomaz+test12@tomaz.me

$ limacharlie users invite --file=1.txt
Found 3 emails in the file.
Inviting user with email tomaz+test1@tomaz.me.
User with email tomaz+test1@tomaz.me already exists / has already been invited.
Inviting user with email tomaz+test2@tomaz.me.
User with email tomaz+test2@tomaz.me already exists / has already been invited.
Inviting user with email tomaz+test12@tomaz.me.
User with email tomaz+test12@tomaz.me has been invited.
```

## Tests

I wanted to add some basic tests, but I noticed our tests work with the actual API (aka they are end to end tests which use real credentials from a testing account).

[...]

PR #156, opened 2025-02-19
fix: [refractionPOINT/tracking#3033] Fix a small potential race condition when writing a config file

tomaz-lc:
## Description

This pull request fixes a small possible race conditions / potential security issue when writing credentials to a file on disk when using a `limacharlie login` function.

Previous code wrote content to a file + then changed the ownership and permission bits. Since those operations are not atomic, there is a possible small time window / race condition after the file has been written and before the ownership + permission bits have been set.

To mitigate this issue, I introduced a new approach where I do the following:

1. Securely generate temporary random file on disk (`mkstemp`)
2. Set ownership + permission bits for this file
3. Write config content to this file
4. Move the temporary file to a final place on disk

rename operation on UNIX is atomic (https://pubs.opengroup.org/onlinepubs/9699919799/functions/rename.html), so this should avoid all possible issues (except perhaps the symlink one, but if we want to handle that, we need to decide how to do it - see code comments).

## Other Changes

While working on this fix, I noticed we already have integration / end to end tests which work with the actual LimaCharlie API and testing credentials, but we don't yet have a concept for unit tests.

As such, I introduced a concept of unit tests which don't require credentials to be set (those live in `tests/unit` directory) and moved existing tests + conftest.py to `tests/integration` directory. Unit tests run as part of PR Cloud Build workflow.

Additionally, I refactored some of the existing code to use more dependency injection to make testing easier.

## Type of change

## TODO

- [x] Add more unit tests for edge cases

## Related issues

Fix refractionPOINT/tracking#3033


maximelb (approved):

tomaz-lc:
One of the integration tests failed (https://console.cloud.google.com/cloud-build/builds/5ab6cf9e-4283-4fe2-b2a1-c957b1dd61de;step=summary?inv=1&invt=AbqAmg&project=legion-212720). It appears to be a flaky test / race conditions since I saw it failing before.

I will retry the build + merge when it passes.

tomaz-lc:
Just a heads up - this PR introduced an inadvertent regression with action args handling in some scenarios.

I'm already working on a fix as part of #155. I will also add some very basic sanity tests which would have caught an issue like that.

tomaz-lc:
I pushed a commit in #155 which should address the regression - c123c0e364c5f82b5b989c831e44c730d7c8cca5.

I also added some very basic high level sanity tests - 35c6e905cfdef429bf0b33e5fb93dafebcb14203.

Sadly those are not the most robust and won't catch many issues, but it's hard to write something general which is more robust due to how our argument parsing works (it's not consistent across different commands / actions).

PR #157, opened 2025-02-21
chore: Bump version in setup.py

tomaz-lc:
## Description

This bumps version in setup.py (something I forgot to do in #155).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 from setuptools import setup
 
-__version__ = "4.9.11"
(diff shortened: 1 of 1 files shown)


tomaz-lc:
@maximelb Thanks. I actually already bumped it in `limacharlie/__init__.py`, but forgot setup.py (usually projects modify setup.py to read it directly from `__init__.py`, but this requires a small path hackery).

Will open PR next week to do that so we don't need to duplicate it in two places.

PR #158, opened 2025-02-23
Add flag to print only data, useful for quick edits.

maximelb:
## Description of the change

Adding simple flag to output only the record data and not the whole record. This makes quick "edit this and push it back" simpler as it avoids the user needing to remove the record envelope from the data returned.

## Type of change


tomaz-lc (approved):
LGTM.

Probably a good idea to also update the docs (if not already).

PR #159, opened 2025-02-24
chore: [refractionPOINT/tracking#3050] Move from `setup.py` to `pyproject.toml`

tomaz-lc:
## Description

This pull request updates the project packaging story and moves from deprecated `setup.py` approach to `pyproject.toml` approach (with setuptools.build_meta build backend).

Additionally, it includes the following changes:

* Include LICENSE, README.md, CHANGELOG.md file with source distribution.
* Use `README.md` for the project long description (this will be rendered in PyPi package page).
* Use `setuptools-scm` package to automatically read the version from SCM (git). This means that when a tag is available (e.g. stable release), it will automatically use that tag for the version and we don't need to update any files and change the version there
* Declare optional dependencies for tests, build, publish
* Update Dockerfile to work with new approach and use multi stage build
* As part of PR CI, also build + basic sanity check the Docker image

## Notes

In theory, this change should be backward compatible, but the package metadata will change a bit and we will include a couple of more files.

This should not affect the end users in any way. I will also add some more dist tests to CI to confirm that. In fact, I already wanted to do it, but Cloud Build makes it a massive pain if you want to do things such as re-use steps, run with multiple versions of Python (you can get it to work using different approaches, but none of them is really pretty and that robust).

## TODO

- [ ] Once this PR is merged, I will update our internal publishing documentation with the relevant changes.

## Type of change

## Related issues

Fix refractionPOINT/tracking#3050


limacharlie/SpotCheck.py
```diff
@@ -123,7 +123,8 @@ def _performSpotChecks( self ):
                 # after being offline.
                 with self._lock:
                     # If there are no more sensors to check, we can exit.
-                    if 0 == self._pendingReCheck and 0 == len( self._sensorsLeftToCheck ):
+                    # NOTE: len() doesn't work on Queue in Python 3
```
tomaz-lc, line 126: I noticed this code was resulting in this error:

```  Traceback (most recent call last):
    File "/home/tomaz/w/refractionPOINT/python-limacharlie/limacharlie/SpotCheck.py", line 120, in _performSpotChecks
      sensor = self._sensorsLeftToCheck.get_nowait()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/lib/python3.11/queue.py", line 199, in get_nowait
      return self.get(block=False)
             ^^^^^^^^^^^^^^^^^^^^^
    File "/usr/lib/python3.11/queue.py", line 168, in get
      raise Empty
  _queue.Empty
  
  During handling of the above exception, another exception occurred:
  
  Traceback (most recent call last):
    File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
      self.run()
    File "/usr/lib/python3.11/threading.py", line 975, in run
      self._target(*self._args, **self._kwargs)
    File "/home/tomaz/w/refractionPOINT/python-limacharlie/limacharlie/SpotCheck.py", line 126, in _performSpotChecks
      if 0 == self._pendingReCheck and 0 == len( self._sensorsLeftToCheck ):
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  TypeError: object of type 'Queue' has no len()
  
    warnings.warn(pytest.PytestUnhandledThreadExceptionWarning(msg))
```

PR #160, opened 2025-02-24
fix [refractionPOINT/tracking#3025] Fix Docker image build + push workflow

tomaz-lc:
# Description

This pull request should fix the Cloud Build workflow which runs on `master` branch and builds + publishes Docker Image to Docker Hub.

## Other Improvements

* Re-try flaky tests on failures (common issue with integration tests)
* Test build + verify .tar.gz source distribution
* Version Docker images - in addition to `latest` also publish tag which corresponds to the tool version

## Type of change

## Related issues

Fix refractionPOINT/tracking#3051

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
-FROM python:slim
+FROM python:3.12-slim
 ADD . /python-limacharlie
 WORKDIR /python-limacharlie
+RUN python -m pip install "setuptools==75.8.0"
 RUN python ./setup.py install
 
 ENTRYPOINT [ "limacharlie" ]
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,16 +1,28 @@
 steps:
-- name: 'gcr.io/${PROJECT_ID}/pytest_3'
-  args: [ '-v', 'tests', '--oid', '${_OID}', '--key', '${_KEY}' ]
-- name: 'gcr.io/cloud-builders/docker'
+- id: "Run Unit Tests"
+  name: 'gcr.io/${PROJECT_ID}/pytest_3'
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM python:slim
+FROM python:3.12-slim
```
tomaz-lc, line 1: I pinned it to avoid any surprises in case a breaking change is made in a new version.

tomaz-lc:
Since integration tests are quite flaky / race conditions prone, I pushed a change to use `pytest` plugin which auto retries failed tests on failure - 25f012b2651bed5c4e95dd0fc8be7b81d9fee58e.

This should hopefully help mitigate flaky tests a bit.

cloudbuild.yaml
```diff
@@ -1,19 +1,48 @@
 steps:
-- name: 'gcr.io/${PROJECT_ID}/pytest_3'
-  args: [ '-v', 'tests', '--oid', '${_OID}', '--key', '${_KEY}' ]
-- name: 'gcr.io/cloud-builders/docker'
```
tomaz-lc, line 4: I moved from a custom pre-built image to a standard image + install dependencies inside so I can install pytest plugin to retry tests on failure.

I could also update pre-built image to bundle in this dependency, but I couldn't find source / Dockerfile for it anywhere and I'm also not sure how it was pushed (I assume it was done manually).

tomaz-lc:
I pushed the tag (https://github.com/refractionPOINT/python-limacharlie/releases/tag/4.9.13) and verified the workflow is now working correctly - https://hub.docker.com/r/refractionpoint/limacharlie/tags

PR #169, opened 2025-03-29
Api key ip

maximelb:
## Description of the change

Add:
- getOrgInfo
- add allowed_ip_range to addApiKey
- add annotations

Remove the reruns of integration tests as it masked the underlying error, took me a while to figure it out.

## Type of change

### DIFF
--- a/cloudbuild_pr.yaml
+++ b/cloudbuild_pr.yaml
@@ -23,7 +23,7 @@ steps:
       pip install -r requirements-tests.txt
 
       # We use rerun since tests are flaky
-      pytest -s -v --durations=10 --reruns 3 --reruns-delay 10 tests/integration/ --oid=${_OID} --key=${_KEY}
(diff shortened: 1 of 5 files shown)


cloudbuild_pr.yaml
```diff
@@ -23,7 +23,7 @@ steps:
       pip install -r requirements-tests.txt
 
       # We use rerun since tests are flaky
-      pytest -s -v --durations=10 --reruns 3 --reruns-delay 10 tests/integration/ --oid=${_OID} --key=${_KEY}
```
tomaz-lc, line 26: I originally added a re-run since I noticed tests are flaky and manually re-running the build in Cloud Build usually helped in such scenarios.

Is this not desired anymore / has something changed?

maximelb: Should not have that in general.

PR #172, opened 2025-04-17
Emit stderr warning on over-quota.

maximelb:
## Description of the change

Adding a stderr warning when we received a header from the API indicating that the request was over-quota (since we don't currently enforce).

## Type of change


josh-lc (approved):
lgtm

PR #173, opened 2025-04-24
Handle jwt-only auth errors better.

maximelb:
## Description of the change

When using a JWT-only SDK, if we get an auth error, there is no way for us to refresh the JWT since we don't have an API key, so we will never get ourselves out of the auth error. So instead of reporting an error down the line about missing an API key, return the original auth error here.

## Type of change


josh-lc (approved):
lgtm

PR #210, opened 2026-02-23
feat: Add transport-level compression, remove application-level is_compressed

tomaz-lc:
## Summary

- Add `Accept-Encoding: zstd, gzip, deflate` header to all API requests for transparent HTTP transport-level compression
- Improve data field compression handling - endpoints that used `is_compressed=true` (base64-encoded gzip inside JSON) now return plain JSON fields, with compression handled transparently at the transport layer instead
- New `transport_compression` module handles `Accept-Encoding` negotiation and `Content-Encoding` decompression (zstd, gzip, deflate)
- `zstandard` added as a hard dependency with runtime graceful fallback - pre-built wheels available for all major platforms, but if unavailable the SDK still works with gzip/deflate only
- `Client.unwrap()` deprecated but kept for backward compatibility

## Why

The previous `is_compressed=true` approach gzip-compressed individual data fields and base64-encoded them inside JSON responses (`base64(gzip(data))`). This required explicit client-side decompression (`unwrap()`) and prevented efficient transport-level compression since compressing already-compressed data is wasteful.

Transport-level compression is better in every way - it's transparent to the caller, handled by standard HTTP content negotiation, covers the entire response body (not just specific fields), and supports modern algorithms like zstd which offer better ratios and faster decompression than gzip.

## Backward compatibility

This change is fully backward compatible:

- The API server continues to support `is_compressed=true` for older clients that haven't upgraded yet
- `Client.unwrap()` is deprecated but still functional for any external callers using it directly
- Uncompressed responses (no `Content-Encoding` header) pass through unchanged
- No breaking changes to any public API surface - all existing method signatures and return types are preserved

There is a companion API-side PR that adds transport compression support on the server. Old clients sending `is_compressed=true` will continue to work as before - the server skips transport compression for those requests to avoid double compression.

## `zstandard` compatibility

`zstandard` is listed as a hard dependency in `pyproject.toml` so pip installs it automatically. It ships pre-built wheels for all major platforms - no C compiler or build toolchain needed at install time.

| Platform | Architecture | Python 3.10 | Python 3.11 | Python 3.12 | Python 3.13 |
|----------|-------------|:-----------:|:-----------:|:-----------:|:-----------:|
| Linux (glibc) | x86_64 | wheel | wheel | wheel | wheel |
| Linux (glibc) | aarch64 | wheel | wheel | wheel | wheel |
| Linux (musl) | x86_64 | wheel | wheel | wheel | wheel |
| Linux (musl) | aarch64 | wheel | wheel | wheel | wheel |
| macOS | x86_64 | wheel | wheel | wheel | wheel |
| macOS | ARM64 (Apple Silicon) | wheel | wheel | wheel | wheel |
| Windows | x86_64 | wheel | wheel | wheel | wheel |
| Windows | ARM64 | wheel | wheel | wheel | wheel |

**Runtime fallback**: If `zstandard` can't be imported (exotic platform without a wheel or C compiler), the SDK still works - `Accept-Encoding` falls back to `"gzip, deflate"` and all transport decompression continues to work via stdlib `zlib`. The zstd code path is guarded with `try/except ImportError` at module load time.

## Changes

| File | Change |
|------|--------|
| `pyproject.toml` | Add `zstandard>=0.22.0` as hard dependency |
| `limacharlie/transport_compression.py` | **New** - compression negotiation + decompression with zstd fallback |
| `limacharlie/client.py` | Add `Accept-Encoding` header, transport decompression, deprecate `unwrap()` |
| `limacharlie/sdk/sensor.py` | Remove `is_compressed` + `unwrap()` from `get_events`, `get_children_events` |
| `limacharlie/sdk/organization.py` | Remove `is_compressed` + `unwrap()` from `get_detections`, `get_audit_logs`, `get_jobs` |
| `limacharlie/config.py` | Fix `os.chown` crash on Windows (Unix-only API) |
| `.github/workflows/ci.yml` | **New** - cross-platform CI (ubuntu/macos/windows x Python 3.10-3.13) |
| `tests/unit/test_transport_compression.py` | **New** - compression module tests including zstd-unavailable fallback |
| `tests/unit/test_client.py` | Transport compression integration tests |
| `tests/unit/test_config.py` | Fix permission assertion for Windows (NTFS uses ACLs, not Unix mode bits) |
| `tests/unit/test_sdk_sensor.py` | Updated for new response format |
| `tests/unit/test_sdk_organization.py` | Updated for new response format |

## Additional fixes

- **Windows compatibility in `config.py`**: Guard `os.chown()` with `hasattr(os, "chown")` - `os.chown` and `os.getuid` don't exist on Windows
- **Windows test fix in `test_config.py`**: `os.chmod(0o600)` doesn't enforce Unix-style permissions on NTFS - skip permission assertion on Windows
- **New GHA CI workflows**: Matrix of ubuntu/macos/windows x Python 3.10/3.11/3.12/3.13 - runs unit tests, wheel install sanity checks, zstd verification, and a dedicated no-zstd fallback job

## Test plan

- [ ] Cross-platform CI passes (ubuntu/macos/windows x Python 3.10-3.13)
- [ ] Verify `Accept-Encoding: zstd, gzip, deflate` sent on all requests
- [ ] Verify zstd decompression round-trip works end-to-end
- [ ] No-zstd fallback CI job passes (zstandard uninstalled, SDK works with gzip/deflate)
- [ ] Integration test with transport compression enabled server-side
- [ ] Verify `unwrap()` still works for external callers

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,84 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [cli-v2, master]
+  push:
+    branches: [cli-v2, master]
+
+jobs:
+  unit-tests:
+    name: "Tests - Python ${{ matrix.python-version }} / ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install package
+        run: |
+          python -m pip install --upgrade pip
+          pip install ".[dev]"
+
+      - name: Verify zstandard is installed
+        run: python -c "import zstandard; print(f'zstandard {zstandard.__version__} OK')"
+
+      - name: Verify transport compression imports
+        run: |
+          python -c "
+          from limacharlie.transport_compression import ACCEPT_ENCODING, decompress_response
+          assert 'zstd' in ACCEPT_ENCODING, f'zstd missing from: {ACCEPT_ENCODING}'
+          print(f'Accept-Encoding: {ACCEPT_ENCODING}')
+          "
+
+      - name: Run unit tests
+        run: pytest -v --durations=10 tests/unit/
+
+  wheel-sanity:
+    name: "Wheel install - Python ${{ matrix.python-version }} / ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.10", "3.13"]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Build and install wheel
+        run: |
+          python -m pip install --upgrade pip setuptools wheel build
+          python -m build
+          pip install dist/limacharlie-*-py3-none-any.whl
+
+      - name: Verify CLI works
+        run: limacharlie version
+
+      - name: Verify zstandard available after wheel install
+        run: python -c "import zstandard; print(f'zstandard {zstandard.__version__} OK')"
+
+      - name: Verify zstd decompression works end-to-end
+        run: |
+          python -c "
+          import zstandard, json
+          from limacharlie.transport_compression import decompress_response
+          # Round-trip: compress then decompress
+          original = json.dumps({'events': [{'type': 'test'}]}).encode()
+          compressed = zstandard.ZstdCompressor().compress(original)
+          result = decompress_response(compressed, 'zstd')
+          assert result == original, 'zstd round-trip failed'
+          print('zstd round-trip OK')
+          "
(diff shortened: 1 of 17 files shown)


maximelb:
Let's add it to the new SDK/CLI that way we can focus our efrorts on what we need to test anyway.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [cli-v2, master]
```
dzimine-lc, line 5: I am -1 on any backward support to old python, we leave it on v1 (v3 to be precise), propose we don't carry any backward compatibiltiy stuff to cli-v2

Suggest you modify it to 1) only act on `cli-v2` branch and remove all backward compatibility part - matrix-python-version.

dzimine-lc (changes requested):
@tomaz-lc the change is good, but it also mixes the compression change with intro of GH action based CI, I recommend we separate the two, and disuss/design/do CI in a diff PR?

.github/workflows/ci.yml
tomaz-lc: I wasn't aware until very recently that `cli-v2` branch will be long running for a while.

Having said that, since this branch is already targeted towards `cli-v2` it should work fine. Once it's merged / takes over the old master, it should work just fine.

In short - it won't run against old version, just against the new one.

tomaz-lc:
@dzimine-lc 

> @tomaz-lc the change is good, but it also mixes the compression change with intro of GH action based CI, I recommend we separate the two, and disuss/design/do CI in a diff PR?

I can split it, yeah, but I needed it to test the changes in this branch + it also detected some existing Windows related bugs :)

Will split CI into a separate branch targeted towards this branch. So the flow is / will be `cli-v2` <- this branch <- new branch with CI changes

dzimine-lc:
consider doing CI changes first.

PR #211, opened 2026-02-24
Add `limacharlie api` raw API command

maximelb:
## Summary

- Adds a generic `limacharlie api <endpoint>` command (like `gh api`) for making authenticated HTTP requests to any LimaCharlie API endpoint
- Adds `Client.raw_request()` method that returns `(status_code, data)` without raising on non-200 responses
- Adds `extra_headers` support to `_rest_call()` and `request()`
- Fixes `ctx.exit()` code propagation in `main()` — with `standalone_mode=False`, Click returns exit codes as return values instead of raising, but `main()` wasn't checking the return value
- Removes replicant-dependent integration tests (superseded by Extensions)
- Fixes CI build issues (version flag, parallel build race condition)

### `limacharlie api` features
- `{oid}` placeholder expansion in endpoint paths
- `-X` / `-f` / `-F` / `--input` for method and body control
- `--target` for hitting alternate API hosts (billing, jwt, stream, etc.)
- `-i` / `--silent` / `--no-auth` / `-H` flags
- Exit codes based on HTTP status (0/4/5)
- Full integration with existing output pipeline (`--output`, `--filter`, `--wide`)

### Test cleanup: replicant phaseout

Replicant services (integrity, exfil, logging, yara, replay) have been superseded by Extensions. Each replicant now has an equivalent extension, making the old `/service/{oid}/{name}` endpoint unnecessary for testing.

**Removed integration tests:**
- `test_yara.py` — YARA source/rule listing via `replicant/yara`
- `test_integrity.py` — Integrity rule CRUD via `replicant/integrity`
- `test_exfil.py` — Exfil watch/event rule CRUD via `replicant/exfil`
- `test_logging.py` — Logging rule CRUD via `replicant/logging`
- `test_replay.py` — Replay dry-run via `replay` service

**Removed from `test_cli_comprehensive.py`:**
- `TestCliIntegrity`, `TestCliLogging`, `TestCliYara`, `TestCliExfil` classes (all made `service_request` calls to replicant services)

**Removed from `conftest.py`:**
- `_REQUIRED_REPLICANTS` list and `_ensure_required_services` fixture that auto-subscribed/unsubscribed replicant services before test runs

**Kept as-is:** SDK classes (`yara.py`, `integrity.py`, `exfil.py`, `logging_rules.py`, `replay.py`) and their unit tests — unit tests mock `service_request` so they pass without live replicant subscriptions, and the SDK classes remain available for users who still have replicant extensions subscribed.

### CI fixes
- `cloudbuild_pr.yaml`: Changed `limacharlie version` → `limacharlie --version` (v2 CLI uses flag, not subcommand)
- `cloudbuild_pr.yaml`: Test steps now copy workspace to `/tmp/src` before `pip install` to avoid a race condition on the `build/` directory when steps run in parallel

## Test plan
- [x] `python -m pytest tests/unit/ -v` — all 712 unit tests pass (including 40 new `test_api_cmd.py` tests)
- [x] CI: 107 integration tests pass, 2 skipped, 0 failures
- [x] CI: wheel and sdist sanity checks pass
- [x] Manual: `limacharlie api orgs/{oid} --output json -i` — returns org info
- [x] Manual: `limacharlie api sensors/{oid} -X GET -f limit=2` — query params work
- [x] Manual: `limacharlie api --target billing orgs/{oid}/status` — billing host alias works
- [x] Manual: `limacharlie api orgs/{oid}/url --no-auth` — public endpoint works
- [x] Manual: exit codes propagate correctly (0 for 200, 4 for 404)


limacharlie/commands/api_cmd.py
```diff
@@ -0,0 +1,268 @@
...
+
+    # --- Expand {oid} placeholder ---
+    if "{oid}" in endpoint:
+        oid = client.oid
+        if oid is None:
+            raise click.UsageError(
+                "Endpoint contains {oid} but no organization ID is configured.\n"
+                "Set a default org with 'limacharlie auth use-org <OID>' or pass --oid."
+            )
+        endpoint = endpoint.replace("{oid}", oid)
```
tomaz-lc: It may not be a bad idea to perform explicit escaping and building of the path components to avoid potential path injections.

limacharlie/commands/api_cmd.py
```diff
@@ -0,0 +1,268 @@
...
+        return (key, True)
+    if raw.lower() == "false":
+        return (key, False)
+
+    # File reference
+    if raw.startswith("@"):
+        path = raw[1:]
+        if path == "-":
+            return (key, sys.stdin.read())
+        with open(path, "r") as f:
```
tomaz-lc: Do we need / want to perform any path validation to mitigate path traversal related security issues (e.g. symlink following, etc.).

I think it is a good idea to add at least some basic safe guards - don't follow symlink, potentially lock paths relative to the cwd / pwd.

limacharlie/commands/api_cmd.py
```diff
@@ -0,0 +1,268 @@
...
+    if raw.lower() == "true":
+        return (key, True)
+    if raw.lower() == "false":
+        return (key, False)
+
+    # File reference
+    if raw.startswith("@"):
+        path = raw[1:]
+        if path == "-":
+            return (key, sys.stdin.read())
```
tomaz-lc: Since read is blocking in this case, I wonder if we should add a guard to refuse to read from terminal when stdin is a tty (e.g. `sys.stdin.isatty()`).

The question / idea is if we need to support piping only or also passing input interactive via stdin? If the former, tty check is a good idea, if the latter, we don't want it (but then there are more edge cases and user may get confusing why the command is hanging - in that case, we should probably print text to the terminal if human friendly format is chosen - e.g. `Waiting for stdin input for argument {key}. Provide the input via stdin and send EOF to proceed`).

limacharlie/commands/api_cmd.py
```diff
@@ -0,0 +1,268 @@
...
+        # Raw body from file or stdin
+        if input_file == "-":
+            body_text = sys.stdin.read()
+        else:
+            with open(input_file, "r") as f:
+                body_text = f.read()
+        raw_body = body_text.encode("utf-8")
+        # Auto-detect content type
+        stripped = body_text.lstrip()
+        if stripped.startswith("{") or stripped.startswith("["):
```
tomaz-lc: Probably also not a bad idea to allow user to explicitly override the content type (aka default to auto, but allow to override via CLI argument).

limacharlie/commands/api_cmd.py
```diff
@@ -0,0 +1,268 @@
...
+            content_type = "application/json"
+        else:
+            content_type = "application/octet-stream"
+    elif fields:
+        if effective_method == "GET":
+            # Fields go to query params for GET
+            query_params = {k: str(v) for k, v in fields.items()}
+        else:
+            # Fields go to JSON body
+            raw_body = json.dumps(fields).encode("utf-8")
```
tomaz-lc, line 281: IIRC, in our API, in many cases we only accept form encoded parameters in the request body (aka in API operations where we don't support JSON, iirc, there are quite a few of those),

limacharlie/commands/api_cmd.py
maximelb: Fixed in bb96526 — the OID value is now URL-encoded via `urllib.parse.quote(oid, safe="")` before substitution into the endpoint path.

limacharlie/commands/api_cmd.py
maximelb: Fixed in bb96526 — new `_safe_read_file()` helper resolves symlinks via `os.path.realpath()` and rejects anything that is not a regular file (directories, broken symlinks, etc.). Used for both `@file` in `-F` and `--input`.

limacharlie/commands/api_cmd.py
maximelb: Fixed in bb96526 — new `_read_stdin()` helper checks `sys.stdin.isatty()` and prints guidance to stderr before blocking: "Reading from terminal stdin. Provide input and press Ctrl-D (EOF) to continue, or Ctrl-C to cancel." Used for both `@-` in `-F` and `--input -`.

limacharlie/commands/api_cmd.py
maximelb: Fixed in bb96526 — new `--content-type` flag for explicit control over `--input` body content type. Validates that it can only be used with `--input` (since field encoding is controlled by `--json` instead).

limacharlie/commands/api_cmd.py
maximelb: Fixed in bb96526 — fields from `-f`/`-F` now default to form-encoded body (passed via the `params` path in `_rest_call`, matching what every other SDK method does). New `--json` flag switches to JSON encoding instead, preserving typed `-F` coercion for bools/ints. This way the default matches LC API conventions while JSON is still available when needed.

maximelb:
@tomaz-lc All 5 review comments addressed in bb96526:

1. **OID path injection** — URL-encoded via `urllib.parse.quote(oid, safe="")`
2. **`@file` path traversal** — `_safe_read_file()` resolves symlinks, rejects non-regular files
3. **stdin TTY guard** — `_read_stdin()` checks `isatty()`, prints guidance before blocking
4. **Content-type override** — new `--content-type` flag (for `--input` only)
5. **Form-encoded default** — fields now default to form-encoded body (matching LC API conventions); new `--json` flag for JSON encoding

48 unit tests pass (8 new for the added behaviors).

tomaz-lc (dismissed):

dzimine-lc (dismissed):

PR #257, opened 2026-03-18
feat: consolidate config files into ~/.limacharlie.d/ directory

tomaz-lc:
## Details

Consolidates all CLI config files (credentials, JWT cache, search checkpoints) under a single directory with platform-appropriate defaults. With CLI v2 (5.x) being a major release, this is the right time to establish consistent file layout conventions and follow platform best practices - rather than shipping with scattered dotfiles (`~/.limacharlie`, `~/.limacharlie_jwt_cache`, `~/.limacharlie.d/`) and accumulating tech debt that would be harder to change post-release.

The new layout unifies everything under `~/.limacharlie.d/` on Unix and `%APPDATA%/limacharlie/` on Windows, with a single `paths.py` module as the source of truth for all path resolution. Full backward compatibility is preserved - existing `~/.limacharlie` configs are auto-detected with a `UserWarning`, and users can migrate via `limacharlie config migrate`.

Migration is intentionally not automatic (consistent with Docker, AWS CLI, Terraform conventions). The CLI reads from the legacy location and emits a warning; users control when migration happens.

### New files
- **`limacharlie/paths.py`** - Centralized path resolution module. Single source of truth for config file, JWT cache, and checkpoint directory paths. Handles platform detection (Unix vs Windows `%APPDATA%`), env var overrides (`LC_CONFIG_DIR`, `LC_CREDS_FILE`, `LC_LEGACY_CONFIG`, `LC_NO_MIGRATION_WARNING`), new-then-legacy fallback with warnings, and per-process caching.
- **`limacharlie/commands/config_cmd.py`** - New `config` CLI command group:
  - `config show-paths` - Display all resolved paths, existence flags, active env var overrides
  - `config migrate` - Copy files from legacy to new layout (`--dry-run`, `--remove-old`, `--force`). Content verification before legacy file removal prevents data loss when destination has different content.

### Modified files
- **`config.py`** - Delegates path resolution to `paths.py`, ensures parent directory exists on write
- **`jwt_cache.py`** - Delegates cache path to `paths.py`
- **`search_checkpoint.py`** - Delegates checkpoint dir to `paths.py`
- **`constants.py`** - Removed duplicate `CONFIG_FILE_PATH`
- **`commands/auth.py`**, **`help_topics.py`** - Updated path references in help text
- **`doc/authentication.md`** - Added config migration guide, JWT caching docs
- **`README.md`** - Added documentation section with links
[...]

PR #254, opened 2026-03-17
feat: streaming output, orjson, and memory-efficient search rendering

tomaz-lc:
## Details

Adds streaming output support across all search command paths, replacing the previous approach that buffered all results in memory before rendering. Output formats that support it (JSONL, JSON, expand, table) now stream results with constant or near-constant memory usage, making large searches (100K+ events) feasible on memory-constrained VMs.

Also adds `orjson` as an optional-but-preferred JSON backend (~3-10x faster than stdlib `json`), with automatic fallback to stdlib if orjson is unavailable.

### Output behavior by path and format

| Path | JSONL | JSON | expand | table | CSV/YAML |
|---|---|---|---|---|---|
| `search run` (no checkpoint) | Stream, O(1) | Stream, O(1) | Stream, O(1) | Stream (sampled widths), O(sample) | Buffer, O(N) |
| `search run --checkpoint` | Stream from file, O(1) | Stream from file, O(1) | Stream from file, O(1) | 2-pass stream from file, O(cols) | Buffer from file, O(N) |
| `search run --resume` | Stream from file, O(1) | Stream from file, O(1) | Stream from file, O(1) | 2-pass stream from file, O(cols) | Buffer from file, O(N) |
| `search checkpoint-show` | Stream from file, O(1) | Stream from file, O(1) | Stream from file, O(1) | 2-pass stream from file, O(cols) | Buffer from file, O(N) |
| `search saved-run` | Stream, O(1) | Stream, O(1) | Stream, O(1) | Stream (sampled widths), O(sample) | Buffer, O(N) |

### Peak memory by format

| Format | Without checkpoint | With checkpoint |
|---|---|---|
| **JSONL** | O(1) - one result at a time | O(1) - one line at a time |
| **JSON** | O(1) - streaming `[`, items, `]` | O(1) - streaming from file |
| **expand** | O(1) - one event block at a time | O(1) - one event block at a time |
| **table** | O(sample_pages * cols) - buffers ~3 pages for column width sampling, then streams | O(cols) - two-pass file scan: pass 1 computes exact column widths, pass 2 streams rows |
| **CSV** | O(N) - inherent to format | O(N) - loaded from file |
| **YAML** | O(N) - inherent to format | O(N) - loaded from file |

### Memory profile - live validation

Measured during a real 30-day `--resume --checkpoint` search over a production org. The checkpoint file grew to 632 MB (484 result pages) while RSS stayed flat at ~138 MB, confirming O(1) memory for the checkpoint write path.

[...]

PR #290, opened 2026-05-07
Merge cli-v2 into master (cli-v2 as source of truth)

tomaz-lc:
## Details

This PR merges the long-running `cli-v2` branch into `master`, with `cli-v2` taken as the source of truth for the final tree. Both histories are preserved: the resulting merge commit has two parents (the previous `master` tip and the `cli-v2` tip), so all 121 cli-v2 commits become reachable from `master`, while the 3 master-only commits since the divergence point (`0eff5c1`) remain in history but their content is superseded.

The merge was constructed locally with the standard "ours-strategy + tree-replace" recipe so the resulting tree exactly equals `cli-v2`'s tree, regardless of conflicts:

```
git merge --no-commit -s ours cli-v2
git read-tree -u --reset cli-v2
git commit
```

Verification: `git diff HEAD cli-v2 --stat` is empty for the merge commit, confirming tree-level equality with `cli-v2`. You can confirm the same on GitHub via the compare view: [cli-v2...merge-cli-v2-into-master](https://github.com/refractionPOINT/python-limacharlie/compare/cli-v2...merge-cli-v2-into-master) - the only file diff visible is the small CI-only follow-up commit described in the next section.

## Additional commit: CI flakiness workaround

On top of the merge commit, this branch also includes one small follow-up commit that hardens `cloudbuild_pr.yaml` against a recurring transient PyPI failure that has been hitting this build at ~40% rate (also affects `cli-v2` itself, not introduced by this merge). The Cloud Build PR check kept failing with `Cache entry deserialization failed, entry ignored` followed by `Skipping page https://pypi.org/simple/<pkg>/ because the GET request got Content-Type: Unknown`, which are well-documented PyPI/CDN flakes (see [pypa/pip#11340](https://github.com/pypa/pip/issues/11340), [pypi/warehouse#12088](https://github.com/pypi/warehouse/issues/12088), [pex-tool/pex#2163](https://github.com/pex-tool/pex/issues/2163)).

The follow-up commit adds the following pip env vars at the Cloud Build `options` level so they apply to all parallel steps and are inherited by `python -m build`'s isolated env: `PIP_NO_CACHE_DIR=1`, `PIP_RETRIES=10`, `PIP_TIMEOUT=60`, `PIP_DISABLE_PIP_VERSION_CHECK=1`. Full rationale is in the inline yaml comment. We do not lose anything meaningful by disabling the pip cache here because Cloud Build does not persist `/root/.cache/pip` across builds anyway. This same fix should be ported back to `cli-v2` (or just inherited from this merge) so future PRs stop hitting the same flake.

## Why this shape

We want `cli-v2` to become the canonical CLI going forward, but we also want to preserve the existing `master` history (commit hashes, PR references, blame). A force-push of `cli-v2` onto `master` would have rewritten history; a plain `git merge cli-v2` would have produced conflicts and a mixed tree. The recipe above is the only approach that gives both: linear history is preserved, and the post-merge tree is byte-identical to `cli-v2`.

## Blast radius / isolation

- Affects: the entire repository tree on `master` is replaced with `cli-v2`'s tree. Every CLI command, SDK module, doc, workflow, and packaging file is touched.
- Not affected: existing tags, other branches, and the `cli-v2` branch itself are untouched. Released PyPI artifacts are unaffected (no release is cut by this PR).
- The 3 master-only commits since divergence (`#280`, `#277`, `#258`) remain reachable in history but their content is overridden by the cli-v2 versions of the same areas (notably `#281` is cli-v2's parallel billing-endpoint migration).

## Notable contracts / APIs

- CLI surface: `cli-v2` is a substantial CLI rework (lazy command loading, streaming output, new commands for cases / feedback / vulnerability / ai-skill / ai-memory / hive schema / etc., new TOON output format, on-disk JWT cache, search retry/checkpoint, consolidated config under `~/.limacharlie.d/`). Users upgrading from the prior `master`-based release will see new commands and behavior changes from the accumulated cli-v2 work.
- Config layout: `cli-v2` consolidates config files into `~/.limacharlie.d/` (see #257). Existing users may need to migrate.
- Packaging / CI: `cloudbuild.yaml` and `Dockerfile` are removed; new GitHub Actions workflows (`ci.yml`, `publish-to-pypi.yml`) are added. Release pipeline expectations change accordingly.

## Performance characteristics

No incremental performance impact from this merge itself (it is a tree replacement). The cli-v2 work it brings in includes streaming output and orjson-based rendering for search (#254), lazy command loading for faster CLI startup (#261), and on-disk JWT caching (#251), all of which improve CLI startup and large-result rendering vs the prior master baseline.

## Review guidance

There is no point reviewing the 121-commit diff line by line in this PR; each of those PRs was reviewed individually on `cli-v2`. The thing to verify here is the merge shape:

- The merge commit has two parents: the previous `master` tip and the `cli-v2` tip.
- `git diff <merge-commit> cli-v2` is empty (tree equals `cli-v2`).
- Master-only commits (`#280`, `#277`, `#258`) are still reachable from `master` history.

## Rollback

If we need to undo: revert the merge commit on `master` (`git revert -m 1 <merge-sha>`). The `cli-v2` branch is untouched and can be re-merged later.


dzimine-lc (commented):
👍

lcbill (approved):

PR #308, opened 2026-06-20
Add CLI support for the new app and ai_cost_model hives

maximelb:
## What was asked

Review the last couple of weeks of changes to the backend hive config schema and determine whether the CLI tool in this repo needs corresponding changes.

## What was done

Two hives were recently added on the backend:

- **`app`** — AI-generated iframe web apps.
- **`ai_cost_model`** — per-org AI cost/savings economic model (cost profiles).

The CLI was only partially caught up:

- `app` already had a shortcut command (added during the cli-v2 sync) but was **missing from the hive discoverability lists**.
- `ai_cost_model` had **no CLI presence at all**.

Changes:

- **`limacharlie/commands/hive.py`** — add `app` and `ai_cost_model` to `_KNOWN_HIVE_TYPES` and to the `hive list` explain text, so both are discoverable via the generic `hive` command.
- **`limacharlie/commands/ai_cost_model.py`** (new) — `ai-cost-model` shortcut command built on the `make_hive_group` factory, bringing it to parity with the other config-like hive shortcuts (`sop`, `note`, `app`, …). The `set` help is field-accurate to the backend record (`label`, `loaded_hourly_rate`, `minutes_per_investigation`, `rate_source_note`).
- **`limacharlie/cli.py`** — register `ai-cost-model` in the lazy-loading command map.
- **`limacharlie/help_topics.py`** — list `app` and `ai-cost-model` wrappers in the `hive` help topic.
- **`tests/unit/test_cli_lazy_loading_regression.py`** — update the CLI-surface snapshot (top-level commands, module map, subcommands, shortcut-load loop).

### Intentionally out of scope

The IaC sync set (`Configs.ALL_HIVES` and the sync `_HIVE_FLAG_MAP`) is left unchanged. It is a curated subset tied to backend `ext-infrastructure` support and already excludes several existing hives (`app`, `sop`, `org_notes`, …), so adding these there would require confirming backend sync support first.

## Testing

Run with the repo `.venv`:

- `pytest tests/unit/test_cli_lazy_loading_regression.py tests/unit/test_discovery.py` → **1003 passed, 2 skipped** (skips are pre-existing, environment-conditional).
- Functional smoke test via `CliRunner`: `ai-cost-model --help`, `ai-cost-model set --help`, and `help topic hive` all render correctly.
- The full `tests/unit` run shows ~111 pre-existing failures unrelated to this change — all in `test_search_*`, caused by the `toon_format` package shipping a `NotImplementedError` stub in this environment. None touch the code changed here.

### DIFF
--- a/limacharlie/commands/ai_cost_model.py
+++ b/limacharlie/commands/ai_cost_model.py
@@ -34,7 +34,7 @@ The data payload describes one costing profile:
     rate_source_note: "FY26 loaded SOC cost / 1,600 hrs"  # optional provenance
 
 loaded_hourly_rate and minutes_per_investigation are required and must
-be non-negative; the backend rejects values outside the allowed range.
+be non-negative; the backend rejects either value above 100000.
 
 Profiles can be tagged and organized with usr_mtd:
 
--- a/limacharlie/commands/ai_cost_model.py
+++ b/limacharlie/commands/ai_cost_model.py
@@ -1,0 +1,57 @@
+"""AI cost model commands for LimaCharlie CLI v2."""
+
+from __future__ import annotations
+
+from ._hive_shortcut import make_hive_group
+from ..discovery import register_explain
+
+group = make_hive_group("ai-cost-model", "ai_cost_model", "AI cost model")
+
+# Override the generic hive explains with cost-model-specific documentation.
+
+register_explain("ai-cost-model.list", """\
+List all AI cost-model profiles stored in the organization.  Each
+profile turns AI-agent activity into an auditable cost/savings figure
+by pairing a fully-burdened analyst rate with the standard manual
+handling time for one investigation of that kind of work.  Because
+this hive is per-org (OID-partitioned), the profiles are per-tenant.
+""")
+
+register_explain("ai-cost-model.get", """\
+Get a specific AI cost-model profile by key.  Returns the full record
+including the economic inputs and metadata.
+""")
+
+register_explain("ai-cost-model.set", """\
+Create or update an AI cost-model profile.
+
+The data payload describes one costing profile:
+
+  data:
+    label: "SOC L1 Triage"           # optional display name
+    loaded_hourly_rate: 95.0         # required: fully-burdened USD/hour
+    minutes_per_investigation: 30.0  # required: standard manual minutes
+    rate_source_note: "FY26 loaded SOC cost / 1,600 hrs"  # optional provenance
+
+loaded_hourly_rate and minutes_per_investigation are required and must
+be non-negative; the backend rejects values outside the allowed range.
+
+Profiles can be tagged and organized with usr_mtd:
+
+  data:
+    label: "SOC L1 Triage"
+    loaded_hourly_rate: 95.0
+    minutes_per_investigation: 30.0
+  usr_mtd:
+    tags: [soc, triage]
+    comment: "FY26 baseline"
+
+Provide data via --input-file (YAML/JSON) or pipe through stdin.
+
+Examples:
+  limacharlie ai-cost-model set --key soc-l1 --input-file model.yaml
+""")
+
+register_explain("ai-cost-model.delete", """\
+Delete an AI cost-model profile.  Requires --confirm.
+""")
--- a/limacharlie/commands/hive.py
+++ b/limacharlie/commands/hive.py
@@ -100,6 +100,8 @@ _KNOWN_HIVE_TYPES = [
     "external_adapter",
     "sop",
     "org_notes",
+    "app",
+    "ai_cost_model",
 ]
 
 
@@ -141,6 +143,8 @@ hold configuration data.  Common hive names include:
   ai_agent         - AI agent configurations
   ai_skill         - Claude Code skill definitions
   ai_memory        - AI agent memories (partial-merge updates)
+  ai_cost_model    - Per-org AI cost/savings economic model
+  app              - AI-generated iframe web apps
 
 Each record returned contains:
   data     - The record payload (structure varies by hive type)
--- a/limacharlie/cli.py
+++ b/limacharlie/cli.py
@@ -98,6 +98,7 @@ def _config_no_warnings() -> bool:
 # The regression test TestModuleMapping verifies this stays in sync.
 _COMMAND_MODULE_MAP: dict[str, tuple[str, str]] = {
     "ai": ("ai", "group"),
+    "ai-cost-model": ("ai_cost_model", "group"),
     "ai-memory": ("ai_memory", "group"),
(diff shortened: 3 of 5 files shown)


lcbill (dismissed):

limacharlie/commands/ai_cost_model.py
```diff
@@ -0,0 +1,57 @@
...
+The data payload describes one costing profile:
+
+  data:
+    label: "SOC L1 Triage"           # optional display name
+    loaded_hourly_rate: 95.0         # required: fully-burdened USD/hour
+    minutes_per_investigation: 30.0  # required: standard manual minutes
+    rate_source_note: "FY26 loaded SOC cost / 1,600 hrs"  # optional provenance
+
+loaded_hourly_rate and minutes_per_investigation are required and must
+be non-negative; the backend rejects values outside the allowed range.
```
maximelb: Help text is field-accurate against the backend record definition: `label` (optional), `loaded_hourly_rate` (required, non-negative, capped), `minutes_per_investigation` (required, non-negative, capped), `rate_source_note` (optional). Good. One nit: the backend enforces upper bounds (both capped at 100000) and rejects values outside the range, but the help only says "non-negative" and "rejects values outside the allowed range" without naming the ceiling. Minor — fine to leave as-is since the backend is the source of truth, but stating the cap would make the `set` help fully self-describing.

limacharlie/commands/ai_cost_model.py
```diff
@@ -0,0 +1,57 @@
+"""AI cost model commands for LimaCharlie CLI v2."""
+
+from __future__ import annotations
+
+from ._hive_shortcut import make_hive_group
+from ..discovery import register_explain
+
+group = make_hive_group("ai-cost-model", "ai_cost_model", "AI cost model")
```
maximelb, line 8: Consistency check: this module overrides the `list`/`get`/`set`/`delete` explains but (correctly) lets `enable`/`disable`/`tag.*` fall back to the generic factory text — same pattern as `app.py`, `sop.py`, `note.py`. Confirmed at runtime: `ai-cost-model.enable` resolves to the generic "Enable an AI cost model by key..." and the overrides for list/set/delete take effect (the module-level `register_explain` calls run after `make_hive_group`, which registers the generics first). No issue.

Also confirmed `--value` is correctly absent from `set` (no `value_key` passed, so this structured-data hive doesn't get the misleading scalar flag), and `make_hive_group(... "AI cost model")` pluralizes to "AI cost models" via the default — `--help` renders cleanly.

limacharlie/commands/hive.py
```diff
@@ -100,6 +100,8 @@ def _record_from_input(key: str, data: Any) -> HiveRecord:
     "external_adapter",
     "sop",
     "org_notes",
+    "app",
+    "ai_cost_model",
```
maximelb, line 104: Completeness note (not a blocker for this PR's scope). I cross-checked the CLI's `_KNOWN_HIVE_TYPES` against the authoritative set of backend `HiveDefinition` names. After this PR, `app` and `ai_cost_model` are covered. The following real backend hives are still NOT in this list (pre-existing gaps, not introduced here): `extension_definition`, `extension_subscription`, `investigation`. If the intent is "CLI fully caught up to backend hives," a follow-up could add those three. Scope of this PR (app + ai_cost_model) is satisfied.

maximelb (commented):
## Adversarial review summary — PR #308

I reviewed the diff, compared against the existing hive-shortcut pattern (`_hive_shortcut.py`, `sop.py`, `note.py`, `app.py`), verified the new help text against the backend record definition, ran the regression test, and did a public-repo hygiene pass.

**Verdict: correct and well-scoped. No defects found.** Findings are minor/informational only.

### Verified
- **Field accuracy**: `ai-cost-model set` help matches the backend record exactly — `label` (optional), `loaded_hourly_rate` (required), `minutes_per_investigation` (required), `rate_source_note` (optional). Required/optional and non-negative constraints all align.
- **Pattern consistency**: built on `make_hive_group` like the other shortcuts; overrides list/get/set/delete explains, falls back to generic enable/disable/tag — identical to `app.py`/`sop.py`/`note.py`. `--value` correctly absent (structured data). Pluralization renders cleanly ("AI cost models").
- **Wiring**: `cli.py` `_COMMAND_MODULE_MAP` entry, `hive.py` `_KNOWN_HIVE_TYPES` + explain text, and `help_topics.py` all updated. `app` (which was already a registered shortcut but missing from discovery lists) is now also added to the discovery lists.
- **Regression test**: snapshot updates are complete and consistent across all four surfaces (top-level commands, module map, subcommands, shortcut load loop). `pytest tests/unit/test_cli_lazy_loading_regression.py` → **993 passed, 2 skipped** (skips pre-existing/env-conditional). Verified at runtime that explain overrides take effect and `--help` renders.
- **Out-of-scope claim is justified**: `Configs.ALL_HIVES` / sync `_HIVE_FLAG_MAP` are a curated subset that already excludes `app`, `sop`, `org_notes`, etc., so leaving them unchanged is correct.
- **Public-repo hygiene**: ✅ no private repo name, no customer names, no PII in the diff or commit message. Backend is referred to generically.

### Minor / informational (non-blocking)
1. `set` help omits the backend's upper bounds (both inputs capped at 100000). "non-negative" is stated; the ceiling isn't. Optional polish.
2. Pre-existing gap (not this PR): `extension_definition`, `extension_subscription`, `investigation` are real backend hives still absent from `_KNOWN_HIVE_TYPES`. Possible follow-up if "fully caught up" is the goal.

Not approving via GitHub's button per instructions; this is a comment-only review.

maximelb:
✅ AI-APPROVED

limacharlie/commands/ai_cost_model.py
maximelb: Addressed in 6cd92f6: the set help now states the explicit 100000 upper bound instead of the vaguer wording.

limacharlie/commands/ai_cost_model.py
maximelb: Confirmed intentional — overriding only list/get/set/delete and letting enable/disable/tag fall back to the generic factory explains matches sop.py / note.py / app.py.

limacharlie/commands/hive.py
maximelb: Deliberately out of scope. extension_subscription/extension_definition are managed via the dedicated 'extension' command (subscribe/config-*), and investigation is runtime incident data rather than hand-edited config, so they are intentionally absent from the generic-hive discoverability list. This PR is scoped to the two new config hives (app, ai_cost_model).

lcbill (approved):

maximelb:
✅ AI-APPROVED

PR #309, opened 2026-06-20
feat(cases): add log-time SDK method + CLI command for human time logging

maximelb:
## What was asked
Humans had no way to log their time on a case from the SDK or CLI. The cases service already exposes a time-logs endpoint, but nothing in this client called it, so the human-effort input to the AI cost/savings model could never be populated. Add the SDK method and CLI command.

## What was done
- **SDK**: `Cases.log_time(case_number, billable_seconds, cost_profile=None, assignee=None, note=None)` → `POST api/v1/cases/{n}/time-logs`, mirroring the existing `add_note` (optional fields omitted from the body when unset).
- **CLI**: `limacharlie case log-time --case-number N --seconds S [--cost-profile K] [--assignee A] [--note T]`, with a client-side positive-seconds guard for a clean error.
- `cost_profile` tags the logged time with a cost-model profile key so it can be valued at that profile's rate downstream.

Key files: `limacharlie/sdk/cases.py`, `limacharlie/commands/case_cmd.py`, plus tests.

## Testing
- `tests/unit/test_sdk_cases.py` — new `TestLogTime` (verb/path/query + body, all optionals, none-excluded).
- `tests/unit/test_cli_case.py` — new `TestCaseLogTime` (minimal, all options, non-positive rejection) + help-list and mocked-method-list updates.
- `tests/unit/test_cli_lazy_loading_regression.py` — added `log-time` to the pinned `case` subcommand set.
- Full unit suite: **3379 passed, 5 skipped** (run with `-o addopts=""` and excluding two tests that need the `toon_format` module, which isn't installed locally — infra-only, unrelated to this change). `pytest-benchmark` is likewise not installed locally; CI has both.

https://claude.ai/code/session_012ie78qt9UBuvCBEEfLRKT3

### DIFF
--- a/limacharlie/commands/case_cmd.py
+++ b/limacharlie/commands/case_cmd.py
@@ -892,6 +892,44 @@ def add_note(ctx, case_number, content, input_file, note_type, is_public) -> Non
 
 
 # ---------------------------------------------------------------------------
+# log-time
+# ---------------------------------------------------------------------------
+
+@group.command("log-time")
+@click.option("--case-number", "case_number", required=True, type=int, help="Case number.")
+@click.option("--seconds", "billable_seconds", required=True, type=int,
+              help="Billable human time spent, in seconds (must be positive).")
+@click.option("--cost-profile", default=None,
+              help="ai_cost_model profile key to value this time at (optional).")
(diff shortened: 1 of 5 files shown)


lcbill (approved):

maximelb:
✅ AI-APPROVED

PR #314, opened 2026-07-17
Bump pytest to 9.0.3 to remediate insecure tmpdir CVE

lcbill:
## Summary

- Remediates Dependabot alert #9: pytest < 9.0.3 relies on a predictable `/tmp/pytest-of-{user}` directory naming pattern on UNIX, which can let a local user cause a denial of service or escalate privileges ([GHSA-6w46-j5rx-g56g](https://github.com/advisories/GHSA-6w46-j5rx-g56g) / CVE-2025-71176, medium severity, CVSS 6.8).
- Closes the gap in the `dev` optional-dependency group in `pyproject.toml`; pytest is a development/test-only dependency and this change has no runtime/production impact.

## Changes

- `pyproject.toml`: bump `pytest` from `==8.3.4` to `==9.0.3` (the first patched version), keeping the existing exact-pin style used for the other `dev` extras.
- Verified `pytest-rerunfailures==15.0` (`pytest !=8.2.2,>=7.4`) and `pytest-benchmark>=5.0.0` (`pytest>=8.1`) both remain compatible with pytest 9.0.3 with no further bumps required.

## Test plan

- [x] Fresh venv, `pip install -e ".[dev]"` — resolves cleanly with pytest 9.0.3, pytest-rerunfailures 15.0, pytest-benchmark 5.2.3.
- [x] Full default suite (`pytest`, i.e. `tests/unit/` + `tests/microbenchmarks/` per `pyproject.toml` `testpaths`): **3534 passed, 5 skipped** (skips are pre-existing/platform-gated: Windows-only, macOS-only, and a documented docstring-coverage exception — unrelated to this bump).
- [x] No test or `conftest.py` changes were needed; nothing in the suite relied on pytest 8-only hooks.
- [ ] `tests/integration/` was not run — it requires live LimaCharlie credentials (`--oid`/`--key` are `required=True` in `tests/integration/conftest.py`) and is excluded from the default `testpaths` already.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,10 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    "pytest==9.0.3",
+    # pytest 9.x (CVE-2025-71176 fix) requires Python >= 3.10; 3.9 keeps the
+    # last compatible release until 3.9 support is dropped
+    "pytest==9.0.3; python_version >= '3.10'",
+    "pytest==8.3.4; python_version < '3.10'",
     "pytest-rerunfailures==15.0",
     "pytest-benchmark>=5.0.0",
     "tomli>=1.0; python_version < '3.11'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    "pytest==8.3.4",
+    "pytest==9.0.3",
     "pytest-rerunfailures==15.0",
     "pytest-benchmark>=5.0.0",
     "tomli>=1.0; python_version < '3.11'",


lcbill:
Fixed the failing Python 3.9 checks: pytest 9.x requires Python >=3.10, so the unconditional `pytest==9.0.3` pin broke resolution on the 3.9 matrix legs (GitHub CI + Cloud Build). The dev extra now uses environment markers — 9.0.3 for >=3.10, and 3.9 keeps the last compatible 8.3.4 (same pattern as the existing `tomli` marker). The CVE-2025-71176 fix is dev-scope (test-runner tmpdir), and the residual 8.3.4 applies only to EOL Python 3.9 dev environments.

maximelb (approved):

PR #320, opened 2026-07-23
Add org description (org_info) support to SDK and CLI

maximelb:
## Summary

Answers a support question: how to set an organization's **description (org info)** via the Python SDK.

The description is only exposed by the backend as an optional side-effect of the org **rename** endpoint (`POST /v1/orgs/{oid}/name`), which requires a non-empty `name`. Previously there was no way to set it from the SDK/CLI — `Organization.rename()` dropped the `description` param and there was no dedicated helper.

## Changes

- **`Organization.rename(new_name, description=None)`** — optionally forwards a `description` to the rename endpoint.
- **`Organization.set_description(description)`** — reads the org's current name via `get_info()` and re-submits it unchanged, so only the description is updated. This is the ergonomic "just set the description on an existing org" call.
- **CLI**:
  - `limacharlie org rename --name ... --description ...`
  - new `limacharlie org set-description --description ...`
- Unit tests (rename with/without description, `set_description` reuse-name + missing-name error) and CHANGELOG entry.

## Notes

- These are thin, honest wrappers over `POST /v1/orgs/{oid}/name` (perm `billing.ctrl`, rate-limited 3/hour server-side).
- The backend skips an empty description and treats the description write as best-effort, so this does not add clear-description semantics the server doesn't support.
- Read the value back via `Organization.get_info()["desc"]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Organization
+
+- **Set org description (org info)**: `Organization.set_description()` /
+  `limacharlie org set-description --description ...` update an organization's
+  description without renaming it (the helper re-submits the current name so
+  only the description changes). `Organization.rename()` and `limacharlie org
+  rename` also take an optional `description` to set both at once. The
+  backend only exposes the description through the rename endpoint, so these
+  are thin wrappers over `POST /v1/orgs/{oid}/name`.
+
 ### Cloud Security (CNAPP)
 
 - **Multi-org fleet overview**: `limacharlie cloudsec fleet overview` /
--- a/limacharlie/commands/org.py
+++ b/limacharlie/commands/org.py
@@ -356,16 +356,41 @@ def delete(ctx: click.Context, confirm_token: str | None) -> None:
 _EXPLAIN_RENAME = """\
 Rename an organization.  The new name must be unique across LimaCharlie.
 Use 'limacharlie org check-name' to verify availability before renaming.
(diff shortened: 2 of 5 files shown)


lcbill (dismissed):

maximelb:
## Auto code review

No blocking issues found. Two independent bug/correctness passes plus manual verification confirmed: `ApiError` resolves, `get_info().get(\"name\")` matches the backend `GET /orgs/{oid}` schema, `query_params` shape is accepted, CLI callbacks wire correctly (verified end-to-end with a mocked org), no regression to existing `rename` callers, and all tests pass (266 unit tests, none skipped).

Two non-blocking notes — both inherent to reusing the rate-limited rename endpoint (the backend only exposes the description as a side-effect of rename), not defects in this code:

1. `set_description()` reads the current name then re-submits it, so a concurrent rename in the (3/hour rate-limited) window between the two calls could be reverted. Acceptable given the backend contract; documented behavior.

https://github.com/refractionPOINT/python-limacharlie/blob/1911bd0bc505a04ae8120a659c66ed78b3583cfe/limacharlie/sdk/organization.py#L231-L239

2. An empty `--description` / `set_description(\"\")` is a silent server-side no-op (the backend skips empty descriptions), so it cannot clear a description and the CLI reports success without warning. Matches the documented backend contract.

https://github.com/refractionPOINT/python-limacharlie/blob/1911bd0bc505a04ae8120a659c66ed78b3583cfe/limacharlie/commands/org.py#L385-L392

lcbill (approved):

PR #325, opened 2026-07-25
fix(deps): make toon_format an optional extra so uv can install current releases

### LINKED ISSUE #324: uv installs silently resolve to 5.3.0 because newer releases depend on pre-release-only toon_format
## Summary

Every `limacharlie` release since 5.3.1 declares `toon_format>=0.9.0b1,<1.0`.
The `toon-format` package has no stable release in that range (PyPI has only
`0.1.0` and `0.9.0b1`), so the constraint is only satisfiable by a pre-release.

`uv` refuses pre-releases of transitive dependencies unless
`--prerelease=allow` is passed. Instead of erroring, its resolver silently
backtracks to the newest version with no pre-release dependency — **5.3.0**
(April 2026). `pip`/`pipx` honor the pre-release specifier per PEP 440 and
install the current release, so the recommended install paths now produce
very different CLIs.

## Reproduction (2026-07-25, uv 0.11.32)

    $ uvx limacharlie --version
    limacharlie, version 5.3.0        # latest on PyPI is 5.5.4

    $ uv tool install limacharlie     # same result: 5.3.0

    $ uv tool upgrade limacharlie
    Nothing to upgrade                # a stranded install self-reports as current

    $ uvx --from 'limacharlie==5.5.4' limacharlie --version
      × No solution found when resolving tool dependencies:
      ╰─▶ Because only toon-format<0.9.0b1 is available and limacharlie>=5.5.4
          depends on toon-format>=0.9.0b1, ...
    hint: `toon-format` was requested with a pre-release marker, but
    pre-releases weren't enabled (try: `--prerelease=allow`)

## Impact

This affects a documented install path: the official docs list
`uv tool install limacharlie` as a supported installation method
(https://docs.limacharlie.io/6-developer-guide/mcp-server/#installation)
with no `--prerelease` flag or minimum-version note, so users following
the docs today get 5.3.0.

This has already bitten downstream: the lc-essentials Claude Code plugin's
CLI auto-install was stranded on 5.3.0 the same way (refractionPOINT/lc-ai#111),
worked around in refractionPOINT/lc-ai#112 by forcing `--prerelease=allow` and
a 5.4.0 version floor in the installer. A packaging fix removes the need for
per-consumer workarounds like that one.

Users who installed via `uv` are pinned to 5.3.0 with no error and no upgrade
path, and miss ~3 months of fixes. Notably, 5.3.0 still contains the
`hive enable`/`disable` bug fixed in 5.4.0 (GET `/mtd` returns `"data": {}`,
the round-tripped record then routes `Hive.set()` to the `/data` endpoint with
`data={}`, and enabling any D&R rule fails with the misleading
`lc_error_code:INVALID - edr rules require "event" or "events" at the
detection root` — the validator is rejecting the empty payload, not the stored
rule). That fix never appeared in the CHANGELOG, which made this hard to
diagnose from the outside.

## Suggested fixes

Any one of these resolves it:

1. Make `toon_format` an optional extra (PR attached — the code already
   treats it as optional at runtime).
2. Publish or depend on a stable `toon-format` release.
3. Document that `uv` installs require `--prerelease=allow`.

Note that exact-pinning `==0.9.0b1` would not help: uv only honors
pre-release specifiers on direct requirements, never transitive ones.

Also worth adding the 5.4.0 `get_metadata()`/`hive enable` fix to the
CHANGELOG retroactively so stranded 5.3.0 users can find it.

## Environment

- uv 0.11.32 (Homebrew, aarch64-apple-darwin), macOS 15
- Reproduced against PyPI as of 2026-07-25

Nynir:
Closes #324.

## Problem

`limacharlie` 5.3.1+ declare a hard dependency on `toon_format>=0.9.0b1,<1.0`. On PyPI, `toon-format` has published exactly two versions — `0.1.0` and `0.9.0b1` — so no *stable* release satisfies that range.

pip honours the pre-release specifier and installs 5.5.4 without complaint. uv does not: it accepts pre-releases only for requirements named directly, never for transitive ones. Faced with a `limacharlie` that needs a pre-release it will not take, uv backtracks to the newest release that does not — 5.3.0.

The failure is silent. No error, no warning:

```console
$ uv pip install --no-cache limacharlie
Resolved 21 packages in 424ms
 + limacharlie==5.3.0
```

`uv tool install limacharlie` and `uvx limacharlie` land on the same 5.3.0, and `uv tool upgrade limacharlie` reports "Nothing to upgrade" — so uv users have been pinned to a release predating the dependency, with no signal that anything is wrong. (The docs list `uv tool install limacharlie` as a supported installation method, so this is a documented path: https://docs.limacharlie.io/6-developer-guide/mcp-server/#installation.) Asking for the version explicitly is the only way to surface the real error:

```console
$ uv pip install --no-cache limacharlie==5.5.4
  × No solution found when resolving dependencies:
  ╰─▶ Because only toon-format<0.9.0b1 is available and limacharlie>=5.5.4
      depends on toon-format>=0.9.0b1, we can conclude that limacharlie>=5.5.4
      cannot be used.
```

Pinning the requirement to the exact pre-release (`toon_format==0.9.0b1`) would **not** fix this — uv's rule is about direct vs. transitive requirements, not about the shape of the specifier.

## Fix

Move `toon_format` out of `[project.dependencies]` into a `toon` extra. The code already treats it as optional at runtime — `limacharlie/output.py` imports it in a `try/except ImportError` with a `_toon_format = None` fallback, the same pattern used for `jmespath` and `tabulate` — and TOON is one of six opt-in `--output` formats, not a default. Nothing but `--output toon` touches it.

With the dependency optional, uv installs the current release normally, no flags (shown here resolving a wheel built from this branch):

```console
$ uv pip install --no-cache limacharlie
Resolved 21 packages in 305ms
 + limacharlie==5.5.5
```

## Behavior change

**A default install loses `--output toon`.** The other five formats (`json`, `yaml`, `csv`, `table`, `jsonl`) are unaffected, and `toon` stays in the `--output` choice list so it still validates and tab-completes.

Requesting TOON without the encoder exits 1 with a message naming both install forms — no traceback, matching how `--filter` reports a missing `jmespath`:

```console
$ limacharlie --output toon sensor list
Error: toon_format is required for --output toon. Install with: pip install 'limacharlie[toon]'
With uv: uv tool install limacharlie --with 'toon-format>=0.9.0b1'
```

The uv-specific form is deliberate. `uv pip install 'limacharlie[toon]'` does not work either — the extra's requirement is still transitive, so uv applies the same rule and quietly falls back to 5.3.0, which does not even have a `toon` extra (it warns, then installs the old version anyway). Naming `toon-format` directly makes it a direct requirement, and uv then honours the pre-release specifier. If `toon_format` ever ships a stable release in range, both forms collapse into the obvious one and the extra keeps working unchanged.

## Tests

CI installs `[dev]`, and `[dev]` now pulls in `toon_format`, so all existing TOON coverage runs unchanged. Added:

- `tests/unit/test_packaging.py` — `toon_format` absent from `[project.dependencies]`, present in the `toon` extra, and present in `dev` so CI keeps exercising the path.
- `tests/unit/test_output.py::TestFormatToonMissingExtra` — with `_toon_format` patched to `None`: `format_toon()` and `format_output(fmt="toon")` both name the extra, the message carries the uv form, the CLI exits 1 without a traceback, and the other formats keep working.
- TOON tests needing a real encoder are marked `skipif`, so `pip install -e .` + `pytest` no longer fails at collection without the extra.

Full suite, both ways:

```
# pip install ".[dev]" (CI's install)
3772 passed, 4 skipped

# without the toon extra
3759 passed, 17 skipped
```

Also verified against a built wheel: fresh venv, `uv pip install <wheel>` with no pre-release flags succeeds and does not pull `toon-format`; `limacharlie --version` works; `--output toon` gives the friendly error; adding the encoder restores TOON output.

## Related

refractionPOINT/lc-ai#112 works around this same resolution failure in the lc-essentials plugin's CLI auto-installer (`--prerelease=allow` + a 5.4.0 floor). With `toon_format` optional, that workaround becomes unnecessary for future releases — and its noted trade-off (that `--prerelease=allow` could also admit a pre-release of `limacharlie` itself) goes away.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,21 @@
 
 ### Dependencies & Python support
 
+- **`toon_format` moved to an optional `toon` extra** (was a hard dependency).
+  The only `toon_format` release in our supported range is a pre-release
+  (`0.9.0b1`), and resolvers that refuse transitive pre-releases — uv among
+  them — cannot install *any* limacharlie version that requires it. uv
+  backtracks silently to 5.3.0 instead, so `uv tool install limacharlie` and
+  `uvx limacharlie` have been stuck on the last release predating the
+  dependency, with `uv tool upgrade` reporting "Nothing to upgrade". Pinning
+  the exact pre-release would not help: uv honours pre-release specifiers only
+  on direct requirements, never on transitive ones. **Behavior change:** a
+  default install no longer supports `--output toon`.
+  `pip install 'limacharlie[toon]'` restores it; under uv, name the package
+  directly (`uv tool install limacharlie --with 'toon-format>=0.9.0b1'`),
+  because asking uv for the extra hits the same transitive pre-release rule.
+  The other five output formats are unaffected, and requesting TOON without
+  the encoder now points at both install forms.
 - **`requests` 2.32.3 -> 2.33.0**: picks up the fix for GHSA-gc5v-m9x4-r6x2
   (insecure temp-file reuse in `extract_zipped_paths()`; vulnerable `<2.33.0`).
 - **Minimum Python is now 3.10** (was 3.9). `requests` 2.33.0 requires Python
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ Python SDK and command-line interface for the [LimaCharlie](https://limacharlie.
 
 ```bash
 pip install limacharlie
+
+# Optional: adds the TOON output format (limacharlie --output toon)
+pip install 'limacharlie[toon]'
+uv tool install limacharlie --with 'toon-format>=0.9.0b1'   # uv equivalent
 ```
 
 ```bash
--- a/doc/cli/README.md
+++ b/doc/cli/README.md
@@ -31,6 +31,18 @@ limacharlie sensor list --output table    # Rich table (default for TTY)
 limacharlie sensor list --output jsonl    # Newline-delimited JSON
 ```
 
+`--output toon` needs the optional `toon` extra; the other formats work with a default install.
+
+```bash
+pip install 'limacharlie[toon]'
+```
+
+Under uv, name the package directly instead — `toon_format` only publishes a pre-release, and uv resolves pre-releases for directly named requirements only, so asking it for the extra makes it fall back to an older `limacharlie`:
+
+```bash
+uv tool install limacharlie --with 'toon-format>=0.9.0b1'
+```
+
 ## Filtering with JMESPath
 
 Use `--filter` with a [JMESPath](https://jmespath.org/) expression to extract or transform output. This works with every command and any output format.
--- a/limacharlie/output.py
+++ b/limacharlie/output.py
@@ -181,9 +181,12 @@ def format_toon(data: Any) -> str:
     See https://toonformat.dev for the spec.
     """
     if _toon_format is None:
+        # uv resolves the extra's pre-release only when toon-format is named as
+        # a direct requirement, hence the separate form.
         raise ImportError(
             "toon_format is required for --output toon. "
-            "Install with: pip install toon_format"
+            "Install with: pip install 'limacharlie[toon]'\n"
+            "With uv: uv tool install limacharlie --with 'toon-format>=0.9.0b1'"
         )
     return _toon_format.encode(data)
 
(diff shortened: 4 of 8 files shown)


tomaz-lc:
@Nynir Good catch & thanks for reporting this issue.

The fix has been merged (with some small related improvements on top) as part of #331.

A new version with that fix (5.6.1) has also been pushed to PyPi (https://pypi.org/project/limacharlie/5.6.1/) - please give it a try.

PR #329, opened 2026-07-30
cloudsec: bind the last four routes + the findings owner filter

maximelb:
## What was asked

Close the remaining gaps in the `cloudsec` integration surface: bind the four `/v1/cloudsec/*`
routes the SDK/CLI did not cover (`/findings/causes`, `/ciem/identities`,
`/data-security/stores`, `/free-tier`), and expose the `owner` findings filter the API accepts
but no client could send.

## What was done

**The four unbound routes**, each mirroring the API's actual parameter set (read from the
gateway handlers, not guessed) and following the existing surface's conventions — cursor/limit
paging, `--output` rendering, docstrings, `--ai-help` explain text, tests:

| SDK | CLI |
|---|---|
| `list_finding_causes()` | `cloudsec finding causes` |
| `list_identity_access()` | `cloudsec ciem identities` |
| `list_data_stores()` | `cloudsec data-security stores` |
| `get_free_tier()` | `cloudsec free-tier` |

**The owner filter.** `finding list|facets` and `export findings` take `--owner` (repeatable)
and `--unassigned`; `finding facets` also takes `--owner-pin`. Without this, `finding
set-owner` could assign an owner that no API caller could then filter by — the write existed
and the read did not.

The unassigned bucket **is** an owner value on the wire (the empty string), so both flags ride
one repeatable selector and "mine or nobody's" is a single filter with two values.
`_owner_selector` folds them and returns `None` when neither is given, so an unfiltered read
sends exactly what it sent before. `--owner-pin` is deliberately separate and documented as
*not* a filter: it keeps named owners visible in the `owner` facet, which is capped at the top
50 by count (`owner_truncated` reports drops), and selects no rows.

**Two facet bindings gained their list's selectors.** `ciem facets` and `data-security facets`
previously took none, while their new list siblings take a full cross-filter — and a facet
count that is computed over a different population than the list it labels is a bug waiting to
be believed. Both now share one selector set with their list. Additive: an unfiltered call
emits the same request as before (covered by a test each). The identity/store booleans are
tri-state — absent leaves the dimension unconstrained, which is *not* the same as pinning it
false — so every one defaults to `None` rather than being a plain flag, and `--mfa` is a
three-value choice because `unknown` is not `off`.

Files: `limacharlie/sdk/cloudsec.py`, `limacharlie/commands/cloudsec.py`, tests in
`tests/unit/test_sdk_cloudsec.py` + `tests/unit/test_cli_cloudsec.py` (+ the cloudsec
subcommand snapshot in `test_cli_lazy_loading_regression.py`), and the user docs
(`doc/cli/cloud-security.md`, `doc/sdk/other-classes.md`, `CHANGELOG.md`).

Not done, deliberately: config CRUD stays out of this surface — `cloudsec_*` records are
managed through the generic `hive` commands, as the module header has always stated.

## Testing

- Full suite: **3803 passed, 5 skipped** (`pytest tests/unit/ tests/microbenchmarks/
  --benchmark-disable`), which is what CI runs.
- New tests cover the wire shapes rather than the plumbing: that an empty owner value survives
  as `?owner=` (dropping it would silently widen the read to the whole estate), that no owner
  flags send no constraint at all, that an omitted tri-state sends nothing while an explicit
  false sends `false`, that a pin is not a filter, and that each new route hits the right path
  with the right selectors.
- Live-verified **read-only** against an internal test org (no write verbs): all four routes
  returned their documented shapes, and the owner round-trip agrees end to end — the facet
  reported 59 unassigned plus 1 assigned within a filter scope, `--unassigned` returned only
  unowned rows, the assigned-owner filter returned exactly that one row, and an owner nobody
  holds returned zero, so the selector is applied rather than ignored.

https://claude.ai/code/session_01L7CkEf1fEa9Y8dBn5Utrx1

### DIFF
--- a/limacharlie/commands/cloudsec.py
+++ b/limacharlie/commands/cloudsec.py
@@ -1,12 +1,15 @@
 """Cloud Security (CNAPP) commands for LimaCharlie CLI v2.
 
 Commands for the ``/cloudsec`` API surface: the merged, risk-ranked
-findings worklist (CSPM misconfigurations + attack paths + CIEM),
-the cloud resource inventory, the pre-aggregated estate topology and
+findings worklist (CSPM misconfigurations + attack paths + CIEM) with
+its facet and shared-fix (cause) rollups, the identity access
+population, the cloud resource inventory, the Data Security (DSPM)
+store list and facets, the pre-aggregated estate topology and
 security graph, compliance assessment, the risk overview, CAASM
 (third-party asset attack surface), sensor<->cloud-asset resolution,
-finding triage, and the cloudsec_policy authoring aids (vocabulary,
-autocomplete, and the "Simulate" matcher previews).
+finding triage, the free-tier standing, and the cloudsec_policy
+authoring aids (vocabulary, autocomplete, and the "Simulate" matcher
+previews).
 
 Reads require the ``cloudsec.get`` permission and writes require
 ``cloudsec.set``. Every command requires the org to be subscribed to
@@ -78,6 +81,25 @@ Examples:
   limacharlie cloudsec scan-status --provider aws
 """
 
+_EXPLAIN_FREE_TIER = """\
+Where the org stands on the cloud-security free tier, and the limits
+that apply to it: is_free_tier, sensor_quota, and — for a free-tier
+org only — max_providers plus enabled_providers ("you are using N of
+M provider connections"). A paying org has no provider cap, so those
+two fields are absent for one.
+
+This endpoint only DESCRIBES the limits; the collector and the
+provider-record validator enforce them, so a limit applies whether or
+not you call this. enabled_providers is omitted rather than zeroed
+when the count could not be read, so an absent field means "unknown",
+never "none configured". There is no trial countdown here — the
+authoritative clock lives in the datacenter, and the collector
+publishes a gated reason through 'cloudsec scan-status'.
+
+Example:
+  limacharlie cloudsec free-tier
+"""
+
 _EXPLAIN_FINDING_LIST = """\
 List the merged, risk-ranked cloud-security findings (CSPM
 misconfigurations + graph toxic-combination attack paths + CIEM
@@ -92,6 +114,10 @@ privilege_escalation, vulnerability, misconfig, coverage_gap,
 device_posture. The canonical, always current list is served by
 'cloudsec finding classes'.
 
+--owner filters by assigned owner (set with 'finding set-owner');
+--unassigned selects the untriaged bucket and combines with --owner,
+so "mine or nobody's" is --owner me@corp.com --unassigned.
+
 Pagination is keyset-based: pass the previous page's next_cursor
 via --cursor.
 
@@ -99,6 +125,8 @@ Examples:
   limacharlie cloudsec finding list --severity CRITICAL
   limacharlie cloudsec finding list --class public_exposure --kev
   limacharlie cloudsec finding list --status open --limit 50
+  limacharlie cloudsec finding list --owner alice@corp.com --status open
+  limacharlie cloudsec finding list --unassigned --severity CRITICAL
 """
 
 _EXPLAIN_FINDING_FACETS = """\
@@ -106,8 +134,39 @@ Cross-filtered facet counts for the findings worklist under the
 same filter selectors as 'finding list'. Each facet dimension is
 counted against the other active filters.
 
-Example:
+The 'owner' facet is keyed by owner, with the empty string holding the
+unassigned bucket, and is capped at the top 50 owners by count —
+'owner_truncated' says whether any were dropped. --owner-pin keeps
+named owners in that capped facet even when they would not rank into
+it (pass your own identity so your row stays reachable on an estate
+with more owners than the cap). It is NOT a filter: it selects no rows
+and changes no count.
+
+Examples:
   limacharlie cloudsec finding facets --severity CRITICAL
+  limacharlie cloudsec finding facets --owner-pin me@corp.com
+"""
+
+_EXPLAIN_FINDING_CAUSES = """\
+Findings grouped by their CAUSE — the mutable object (a firewall
+rule, the principal an attack path's grants land on) whose single
+edit resolves every finding under it. Lets a worklist be worked by
+fix instead of by row: "change this one thing, close N findings".
+
+Takes the same filters as 'finding list', so a rollup can be scoped
+exactly like the list it summarizes. Pass --cause for the exact
+count of one cause; omit it for the top causes by count (--limit,
+default 20, cap 200). 'distinct' is the total number of matching
+causes, so you can tell how much tail the head hides — the counts
+themselves are always exact.
+
+'kind' is an open vocabulary that grows server-side; treat an
+unrecognized value as "some object" rather than assuming a class.
+
+Examples:
+  limacharlie cloudsec finding causes
+  limacharlie cloudsec finding causes --severity CRITICAL --limit 5
+  limacharlie cloudsec finding causes --cause "lcrn:...:firewalls/allow-all"
 """
 
 _EXPLAIN_FINDING_GET = """\
@@ -192,10 +251,50 @@ Example:
 """
 
 _EXPLAIN_CIEM_FACETS = """\
-CIEM identity facet counts (identity kinds, external/public splits).
+CIEM identity facet counts (identity kinds, MFA state, risk bands,
+plus the admin/external/public/disabled/dormant/escalation/
+sensitive-access rollups and the total), so the identity view can
+lead with insight before the row list.
 
-Example:
+The selectors CROSS-FILTER the rail: each dimension is counted under
+the other active selectors but not its own, so a value's count is
+exactly how many rows selecting it would list. With no selectors the
+response is the whole-population rollup. It takes the SAME selectors
+as 'ciem identities', so the counts and that list always describe the
+same population.
+
+--mfa unknown is everyone the MFA question does not apply to (no
+identity-provider observation, or non-human) — it is NOT 'off'. The
+boolean filters are tri-state: omitting one leaves the dimension
+unconstrained, which is not the same as pinning it false.
+
+Examples:
   limacharlie cloudsec ciem facets
+  limacharlie cloudsec ciem facets --kind service_account --admin
+"""
+
+_EXPLAIN_CIEM_IDENTITIES = """\
+One page of the Access screen's identity population: the same
+per-principal effective-access rollup rows 'ciem public-access'
+carries (grant / privileged / sensitive-reach counts, posture facets,
+risk score), but server-filtered and pageable instead of a
+risk-ranked top-N.
+
+Takes the same selectors as 'ciem facets', so the rail's counts and
+this list always describe the same population. Ranked by risk score
+descending by default; a walk that spans a projector recompute can
+move a row across the cursor, so use it for browsing rather than for
+exact exports.
+
+--account / --region are served from the projector's merged view: on a
+tenant whose view has not been built yet the call fails loudly rather
+than returning a misleading zero.
+
+Examples:
+  limacharlie cloudsec ciem identities --limit 50
+  limacharlie cloudsec ciem identities --kind service_account --can-escalate
+  limacharlie cloudsec ciem identities --external --with-sensitive
+  limacharlie cloudsec ciem identities --mfa off --admin
 """
 
 _EXPLAIN_CIEM_IDENTITY = """\
@@ -247,8 +346,30 @@ _EXPLAIN_DATA_SECURITY_FACETS = """\
 DSPM data-store facet counts: total/sensitive/public/public-sensitive
 data stores plus store-kind, sensitivity, and exposure histograms.
 
-Example:
+The selectors cross-filter the rail (each dimension counted under the
+other active selectors but not its own) and are the SAME ones 'stores'
+takes, so the counts always describe the population that list returns.
+
+Examples:
   limacharlie cloudsec data-security facets
+  limacharlie cloudsec data-security facets --store-kind bucket
+"""
+
+_EXPLAIN_DATA_SECURITY_STORES = """\
+One keyset page of the org's data stores, served from the
+materialized graph store under the same selectors as
+'data-security facets' — so the filtered list stays exact at any
+estate size instead of client-filtering a capped walk.
+
+--sensitive / --public are tri-state: omitting one leaves the
+dimension unconstrained; --not-sensitive / --not-public pin it to
+false. Rows are ordered by a stable stored key, so a full walk with
+--cursor is safe.
+
+Examples:
+  limacharlie cloudsec data-security stores --limit 50
+  limacharlie cloudsec data-security stores --sensitive --public
+  limacharlie cloudsec data-security stores --store-kind bucket --data-class pii
 """
 
 _EXPLAIN_RESOURCE_GET = """\
@@ -617,8 +738,10 @@ register_explain("cloudsec.changes", _EXPLAIN_CHANGES)
 register_explain("cloudsec.risk-trend", _EXPLAIN_RISK_TREND)
 register_explain("cloudsec.scan-status", _EXPLAIN_SCAN_STATUS)
 register_explain("cloudsec.topology", _EXPLAIN_TOPOLOGY)
+register_explain("cloudsec.free-tier", _EXPLAIN_FREE_TIER)
 register_explain("cloudsec.finding.list", _EXPLAIN_FINDING_LIST)
 register_explain("cloudsec.finding.facets", _EXPLAIN_FINDING_FACETS)
+register_explain("cloudsec.finding.causes", _EXPLAIN_FINDING_CAUSES)
 register_explain("cloudsec.finding.classes", _EXPLAIN_FINDING_CLASSES)
 register_explain("cloudsec.finding.get", _EXPLAIN_FINDING_GET)
 register_explain("cloudsec.finding.resolve", _EXPLAIN_FINDING_RESOLVE)
@@ -629,9 +752,11 @@ register_explain("cloudsec.attack-path.list", _EXPLAIN_ATTACK_PATH_LIST)
 register_explain("cloudsec.ciem.public-access", _EXPLAIN_CIEM_PUBLIC_ACCESS)
 register_explain("cloudsec.ciem.facets", _EXPLAIN_CIEM_FACETS)
 register_explain("cloudsec.ciem.identity", _EXPLAIN_CIEM_IDENTITY)
+register_explain("cloudsec.ciem.identities", _EXPLAIN_CIEM_IDENTITIES)
 register_explain("cloudsec.inventory.list", _EXPLAIN_INVENTORY_LIST)
 register_explain("cloudsec.inventory.facets", _EXPLAIN_INVENTORY_FACETS)
 register_explain("cloudsec.data-security.facets", _EXPLAIN_DATA_SECURITY_FACETS)
+register_explain("cloudsec.data-security.stores", _EXPLAIN_DATA_SECURITY_STORES)
 register_explain("cloudsec.resource.get", _EXPLAIN_RESOURCE_GET)
 register_explain("cloudsec.graph.neighbors", _EXPLAIN_GRAPH_NEIGHBORS)
 register_explain("cloudsec.query.list", _EXPLAIN_QUERY_LIST)
@@ -816,6 +941,10 @@ _SIMULATE_TARGET_CHOICES = click.Choice(
 _SUGGEST_DIMENSION_CHOICES = click.Choice(
     ["name", "account"], case_sensitive=False,
 )
+# The identity MFA selector is three-state by contract ("unknown" is not
+# "off"), and the set is closed server-side — a Choice keeps a typo from
+# reading as "no MFA constraint".
+_MFA_CHOICES = click.Choice(["on", "off", "unknown"], case_sensitive=False)
 # Provider and CAASM-source values are deliberately NOT click.Choice lists:
 # both are growing server-side registries (new providers/sources ship without
 # a CLI release) and the server rejects unknown values with a clean error.
@@ -833,6 +962,20 @@ def _paging_options(f):
     return f
 
 
+def _owner_selector(owners, unassigned: bool) -> list[str] | None:
+    """Fold --owner/--unassigned into the wire `owner` selector.
+
+    The unassigned bucket IS an owner value on the wire: the empty
+    string. It rides the same repeatable selector, so "mine or
+    nobody's" is one filter with two values. No selection at all
+    returns None so the request carries no owner constraint.
+    """
+    values = list(owners)
+    if unassigned:
+        values.append("")
+    return values or None
+
+
 def _finding_filter_options(f):
     """The findings worklist filter selectors (shared by list/facets).
 
@@ -842,6 +985,16 @@ def _finding_filter_options(f):
     f = click.option(
         "-q", "--search", "q", default=None,
         help="Substring search over the findings.",
+    )(f)
+    f = click.option(
+        "--unassigned", is_flag=True, default=False,
+        help="Include findings with no owner (the untriaged bucket); "
+             "combines with --owner.",
+    )(f)
+    f = click.option(
+        "--owner", "owners", multiple=True,
+        help="Filter by assigned owner; repeatable (OR). Use --unassigned "
+             "for findings nobody owns.",
     )(f)
     f = click.option(
         "--kev/--no-kev", "kev", default=None,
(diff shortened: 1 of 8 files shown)


lcbill (dismissed):

limacharlie/sdk/cloudsec.py
```diff
@@ -248,14 +331,75 @@ def get_finding_facets(
...
+
+        Returns:
+            ``{"causes": [{"key": str, "name": str, "kind": str,
+            "count": int}, ...], "distinct": int}``. ``kind`` is an OPEN
+            vocabulary (``firewall_rule``, ``entitled_identity``, ...) that
+            grows server-side — treat an unrecognized value as "some
+            object" rather than assuming a class. ``distinct`` is the total
+            number of causes matching the filter and MAY exceed
+            ``len(causes)`` when ``limit`` truncates; the counts themselves
+            are always exact.
```
maximelb: Two claims here are stronger than what the API actually delivers.

1. **"returned as a single-entry `causes`"** — when the cause key matches no finding under the filter (it was just fixed, or the key is mistyped/clamped), the backend's `GROUP BY` yields no row and the response is `{"causes": [], "distinct": 0}`. A caller following this docstring with `resp["causes"][0]["count"]` gets an `IndexError`. Please say "a single entry, or empty when nothing matches".

2. **"the counts themselves are always exact"** — the rollup counts on the *stored* status column rather than recomputing the disposition overlay per row, so a **lapsed acceptance** is counted under the status stamped at disposition time while reading that same finding returns `open`. The contrast you're drawing is "exact rather than a capped lower bound", which is true and worth keeping; "always exact" unqualified promises more. Suggest qualifying it here and in `_EXPLAIN_FINDING_CAUSES` (`limacharlie/commands/cloudsec.py:161`), which repeats the phrase verbatim.

limacharlie/commands/cloudsec.py
```diff
@@ -898,6 +1051,125 @@ def _inventory_filter_options(f):
...
+             "— it is NOT 'off'.",
+    )(f)
+    f = click.option(
+        "--risk-band", "risk_bands", multiple=True,
+        help="Filter by risk band (critical/high/medium/low); repeatable (OR).",
+    )(f)
+    f = click.option(
+        "--criticality", "criticalities", multiple=True,
+        help="Filter by crown-jewel tier; repeatable (OR).",
+    )(f)
```
maximelb, line 1151: `--risk-band` and `--criticality` are **closed** vocabularies server-side, but they're free text here, so a typo fails *closed* rather than loudly:

- `risk_band` is matched against a fixed four-token map (`critical|high|medium|low`); an unrecognized token makes the backend add a `FALSE` predicate, so `--risk-band criticl` returns **0 rows with exit 0**.
- `criticality` (and `--tier` on `data-security`) is matched with `IN UNNEST(...)` against the classifier's four tiers, which are the same lower-case `critical|high|medium|low` — again silently empty on a typo.

That is exactly the rationale you wrote for `_MFA_CHOICES` ("the set is closed server-side — a Choice keeps a typo from reading as 'no MFA constraint'"), and unlike `--provider`/`--source` neither of these is a growing server-side registry. Suggest `click.Choice` for `--risk-band`, and at minimum naming the four tiers in the `--criticality` / `--tier` help (today it just says "crown-jewel tier", so a user has nothing to type).

Related evidence that the vocabulary wasn't checked: the new tests pass `--criticality tier1` and `--tier tier1` (`tests/unit/test_cli_cloudsec.py:1179`, `:1253`). `tier1` is not a value the product emits, so those invocations would return nothing live — the mock hides it.

limacharlie/commands/cloudsec.py
```diff
@@ -898,6 +1051,125 @@ def _inventory_filter_options(f):
...
+    f = click.option(
+        "--account", "accounts", multiple=True,
+        help="Filter by account/project; repeatable (OR). Served from the "
+             "merged view, like --region.",
+    )(f)
+    f = click.option(
+        "--source", "sources", multiple=True,
+        help="Filter by producing sweep (okta, gcp, google_workspace, ...); "
+             "repeatable (OR). Alias of --provider.",
```
maximelb: "Alias of `--provider`" points at a flag that does not exist on either command this decorator is applied to (`ciem facets`, `ciem identities`) — only the SDK takes both. Worse, `--provider` *does* exist elsewhere in this group (`inventory list --provider`, `data-security stores --provider`) with different semantics, so the help reads as a cross-reference to something it isn't. Either add `--provider` as a real repeatable alias here (the API unions the two names, so it costs nothing) or drop the sentence / name the API parameter explicitly.

limacharlie/commands/cloudsec.py
```diff
@@ -898,6 +1051,125 @@ def _inventory_filter_options(f):
...
+    )(f)
+    f = click.option(
+        "--public/--not-public", "exposure", default=None,
+        help="The 'exposure' dimension: only publicly-exposed (or only "
+             "non-public) stores. Omit for no constraint.",
+    )(f)
+    f = click.option(
+        "--sensitive/--not-sensitive", "sensitivity", default=None,
+        help="The 'sensitivity' dimension: only sensitive (or only "
+             "non-sensitive) stores. Omit for no constraint.",
```
maximelb, line 1195: The negation convention splits inside one command group: `_identity_filter_options` (added in the same diff) uses `--no-public` / `--no-crown-jewel` / `--no-disabled`, while these use `--not-public` / `--not-sensitive`. Same word, same tri-state, two spellings — so `ciem identities --not-public` and `data-security stores --no-public` are *both* usage errors, and a user who learns one command gets bitten by the other. Pick one form for both new decorators (`doc/cli/cloud-security.md:106` already has to spell out the exception, which is the tell).

limacharlie/commands/cloudsec.py
```diff
@@ -930,12 +1202,13 @@ def group() -> None:
       risk-trend          Risk-score history
       scan-status         Cloud-collection run status per provider
       topology            Pre-aggregated estate topology (exact at scale)
+      free-tier           Free-tier standing and the limits that apply
       fleet overview      Multi-org fleet posture board (MSSP)
-      finding ...         Findings worklist + triage (resolve, owner, ticket, classes)
+      finding ...         Findings worklist + triage (resolve, owner, ticket, causes)
       attack-path list    Headline toxic-combination attack paths
-      ciem ...            Identity access views (public-access, facets, identity)
+      ciem ...            Identity access views (public-access, facets, identities)
```
maximelb: This listing now omits two commands that still exist: `finding classes` (it was named here before, replaced by `causes`) and `ciem identity` (replaced by `identities`). The singular/plural pair is precisely the one that needs both spelled out here, since `ciem identity <urn>` and `ciem identities` are different commands and only the plural is now advertised. Suggest `(resolve, owner, ticket, causes, classes)` and `(public-access, facets, identity, identities)`.

tests/unit/test_cli_cloudsec.py
```diff
@@ -110,7 +111,7 @@ def test_ciem_subgroup_help(self):
         runner = CliRunner()
         result = runner.invoke(cli, ["cloudsec", "ciem", "--help"])
         assert result.exit_code == 0
-        for cmd in ["public-access", "facets", "identity"]:
+        for cmd in ["public-access", "facets", "identity", "identities"]:
```
maximelb: `"identity"` is now a vacuous assertion: it is a substring of `"identities"`, so this loop passes even if the `identity` command is deleted from the group. Anchor it so the singular command keeps real coverage — e.g. assert on the padded command column (`"identity "`) or a `re.search(r"^\s+identity\s", output, re.M)`.

maximelb:
Verified against the gateway handlers and the backend RPC parsers: all four route bindings send exactly the parameters the API forwards, with the right names/shapes, nothing extra, nothing missing; `?owner=` survives `urlencode` to the wire and to the backend filter; the tri-states distinguish absent from false end to end; every gateway `/cloudsec/*` route is now bound. Suite reproduced: 3803 passed, 5 skipped. No customer data, PII, or private repo/identifier leaks.

Open findings:
- `list_finding_causes` docstring promises a single-entry `causes` — a cause matching nothing returns `[]`, so the documented `causes[0]` access raises.
- Same docstring (and `_EXPLAIN_FINDING_CAUSES`) says counts are "always exact"; the rollup counts the stored status, which diverges from the read status for a lapsed acceptance.
- `--risk-band` / `--criticality` / `--tier` are closed four-value server vocabularies left as free text, so a typo returns 0 rows with exit 0 — the same reason `_MFA_CHOICES` exists. Help never names the tier values.
- New tests pass `--criticality tier1` / `--tier tier1`, values the product never emits; the mock hides that the invocation is a no-match live.
- `--source` help says "Alias of --provider", but no `--provider` option exists on `ciem facets` / `ciem identities` (and it means something else on `inventory list`).
- Negation convention split in the same group: `--no-public` (identity filters) vs `--not-public` (data-store filters).
- `group()` help listing dropped `finding classes` and `ciem identity`; only the plural `identities` is now advertised.
- `test_ciem_subgroup_help`: `"identity"` is a substring of `"identities"`, so that assertion can no longer fail.

limacharlie/sdk/cloudsec.py
maximelb: Fixed in d205e40: both claims corrected. The `cause` lookup is now documented as a list to read rather than an element to index (an empty `causes` means no finding under the filter carries it), and "the counts themselves are always exact" is gone — replaced with what that line was actually reaching for (whole-population rather than a capped lower bound) plus the stored-status lag on a self-expired acceptance. Same correction applied to _EXPLAIN_FINDING_CAUSES.

limacharlie/commands/cloudsec.py
maximelb: Fixed in d205e40: confirmed both vocabularies against the backend (identityRiskBandExpr's four-token map + the FALSE fallback, and classifier.KnownTiers()), and all three selectors — --risk-band, --criticality, --tier — now take a click.Choice of critical/high/medium/low. Since that would have removed the ability to select the empty (no-tier) value, --unclassified now does it explicitly on both commands, folded through the same helper as --unassigned. Added tests that a bad band/tier is rejected at parse time, and the tests that used the fictional 'tier1' now use a real tier.

limacharlie/commands/cloudsec.py
maximelb: Fixed in d205e40: the help now names the dimension instead of a flag that isn't on these commands — "Same dimension 'inventory list' calls --provider".

limacharlie/commands/cloudsec.py
maximelb: Fixed in d205e40: unified on --no-sensitive / --no-public, which also matches the pre-existing --no-kev / --no-reachable in this file rather than introducing a third convention. The test that exercised --not-public caught the rename, and docs/changelog were updated with it.

limacharlie/commands/cloudsec.py
maximelb: Fixed in d205e40: both are listed again — 'finding ... (resolve, owner, ticket, causes, classes)' and 'ciem ... (public-access, facets, identities, identity)'.

tests/unit/test_cli_cloudsec.py
maximelb: Fixed in d205e40: the assertion now compares the ciem group's registered command names as a set, so the singular command cannot be deleted without failing — and the comment records why a substring check can't distinguish identity from identities.

lcbill (dismissed):

limacharlie/commands/cloudsec.py
```diff
@@ -1109,27 +1431,69 @@ def finding_list(ctx, severities, finding_classes, statuses, accounts,
...
         reachable=reachable,
         kev=kev,
         q=q,
     ))
 
 
+@finding_group.command("causes")
+@_finding_filter_options
+@click.option("--cause", default=None,
+              help="One cause key: return the exact count for that cause "
```
maximelb: The "always exact" correction did not reach the two places a user is most likely to read. `d205e40` qualified the SDK docstring and `_EXPLAIN_FINDING_CAUSES`, but this `--help` string still says "return the **exact** count for that cause alone", and `CHANGELOG.md:32` still says "`--cause` returns the **exact** count for one cause".

The single-cause path has exactly the same caveat as the rollup — `TenantStore.CauseFacets`' single-key branch is the same `COUNT(*)` over the same `buildFilter`, so it counts the STORED status column and a lapsed acceptance is counted under its stale value until the 6h heal sweep. `--ai-help` now says so; `--help` promises otherwise, and `--help` is what people read.

Suggest matching the explain text here ("the count for that cause alone" / "the whole-population count") and in the changelog bullet.

tests/unit/test_cli_cloudsec.py
```diff
@@ -1062,3 +1070,311 @@ def test_findings_empty_match_allowed(self):
...
+
+class TestClosedVocabularyGuards:
+    """The closed server vocabularies fail CLOSED, so a typo must not parse.
+
+    An unrecognized risk band contributes a FALSE predicate and a
+    misspelled tier matches no row, so without validation a typo would
+    exit 0 with an empty result under a filter the user can see applied.
+    """
+
+    def test_rejects_unknown_risk_band(self):
```
maximelb: These three guards cannot fail, so they do not lock in the Choice they were added for. They run **unmocked** (no `_patches()`) and assert only `exit_code != 0`. Measured on this branch:

- `ciem identities --risk-band urgent` → exit **2**, `Error: Invalid value for '--risk-band'` (what the test intends to catch)
- `ciem identities --risk-band critical`, no credentials (the CI condition) → exit **1**

So with `type=_RISK_BAND_CHOICES` deleted, the invocation would fall through to `_get_cloudsec()`, fail on credentials, exit 1, and all three tests would still pass. The same shape applies to the pre-existing `test_identities_rejects_unknown_mfa_state` (:1229).

Assert something only a parse rejection produces: `assert result.exit_code == 2` and/or `"Invalid value for '--risk-band'" in result.output`. (Alternatively run them under `_patches()` and assert the SDK method was never called.)

limacharlie/sdk/cloudsec.py
```diff
@@ -248,14 +331,86 @@ def get_finding_facets(
         Takes the same filter selectors as :meth:`list_findings`; each
         facet dimension is counted against the other active filters.
 
+        Args:
+            owner_pin: Owners to keep in the ``owner`` facet even when they
```
maximelb, line 336: This states the pin guarantee unconditionally, but the backend's is bounded and the bound is reachable from this SDK. In `findingstore.facetOwner`, `Owners` and `OwnerFacetPins` are concatenated and then clamped to `maxOwnerFacetValues` (50) **combined**, and the unassigned bucket outranks the pins (`ORDER BY unassigned DESC, selected DESC, …`) — so it consumes one of the 50 slots. The gateway allows 100 values per key (`maxCloudSecFilterValues`), so a caller passing e.g. 60 `owner` values plus a pin gets the pin silently dropped, and `owner_truncated` does not distinguish "part of your own selection/pin was dropped" from ordinary tail truncation (the store's own comment says exactly this).

Given the rest of this docstring set is careful about bounds, worth one clause: the pin holds while the active `owner` selection plus pins stays under the 50-value cap; past that, render a selected-but-absent owner as zero rather than trusting the map.

limacharlie/commands/cloudsec.py
```diff
@@ -898,6 +1069,139 @@ def _inventory_filter_options(f):
...
+             "— it is NOT 'off'.",
+    )(f)
+    f = click.option(
+        "--risk-band", "risk_bands", multiple=True, type=_RISK_BAND_CHOICES,
+        help="Filter by risk band; repeatable (OR). The band token the rail "
+             "renders, not a numeric range.",
+    )(f)
+    f = click.option(
+        "--unclassified", "unclassified", is_flag=True, default=False,
+        help="Include identities with no crown-jewel tier assigned; "
```
maximelb: `--unclassified` will be read as the complement of `--no-crown-jewel` (three flags in this one `--help` share the phrase "crown-jewel"), and it isn't. They are different columns and the sets are not complements:

- `--crown-jewel/--no-crown-jewel` → `IFNULL(i.is_sensitive,false) = @if_crown_jewel`
- `--unclassified` → `IFNULL(i.criticality,'') IN UNNEST([""])`

In `go-cloudsec/classifier`, `Classification.Sensitive` is set by ANY matching rule while `Tier` stays `""` unless a matched rule assigned one — so a declared crown jewel whose rule sets no `tier` has `is_sensitive=true, criticality=""`. `--unclassified` returns those too, i.e. it is a strict superset of `--no-crown-jewel`.

Suggest naming the column rather than the concept, e.g. "Include identities with no criticality tier assigned (independent of --crown-jewel: a declared crown jewel whose rule set no tier is unclassified too)". The `data-security` twin already avoids this by saying "criticality tier".

maximelb:
Re-reviewed against `d205e40`. All eight earlier findings independently verified as fixed — the Choice sets are correct and complete for all three flags (`identityRiskBandExpr`'s four-token map for `risk_band`, `classifier.KnownTiers()` for `criticality`/`tier`, which are the same four values), the `--unclassified` escape hatch really does select the no-tier bucket end to end (`IFNULL(i.criticality,'')`/`IFNULL(criticality,'') IN UNNEST` + the gateway's `addCloudSecMultiValue` forwarding a lone `""` as a one-element array), `click.Choice(case_sensitive=False)` normalizes to the lowercase token the backend matches, `_selector_with_empty` is semantically identical to `_owner_selector` with all 8 call sites updated, and no `--not-*` spelling survives anywhere. Suite: 3809 passed, 5 skipped. No customer data, PII, private repo names or internal identifiers in the diff.

Open findings:
- `--cause` `--help` and `CHANGELOG.md:32` still say "the exact count" — the phrase the fix removed from the SDK docstring and `_EXPLAIN_FINDING_CAUSES`; the single-cause branch has the same stored-status caveat.
- The three new closed-vocabulary guard tests are vacuous: unmocked + `exit_code != 0` passes at exit 1 (no credentials) as well as exit 2 (parse rejection), so deleting the Choice would keep them green.
- `owner_pin` docstring guarantee is unconditional; `facetOwner` clamps `Owners`+`OwnerFacetPins` to 50 *combined* with the unassigned bucket taking a slot, and `owner_truncated` can't report a dropped pin.
- `--unclassified` on `ciem` reads as the complement of `--no-crown-jewel` but is a strict superset: `criticality=""` also covers a declared crown jewel whose matching rule set no `tier`.

lcbill (dismissed):

limacharlie/commands/cloudsec.py
maximelb: Fixed in f286f84: --cause's --help and the changelog line now say 'the count' (plus the empty-result case in the help), so --help and --ai-help no longer disagree.

tests/unit/test_cli_cloudsec.py
maximelb: Fixed in f286f84: they now assert exit 2 AND click's "Invalid value for '--flag'", and I verified the fix by mutation — deleting the three Choices fails four tests (the three new guards plus a falsifiable --mfa sibling I added, since the pre-existing one had the same weakness), and restoring them passes all four. Good catch: a guard that cannot fail is worse than no guard, because it reads as coverage.

limacharlie/sdk/cloudsec.py
maximelb: Fixed in f286f84: the docstring now states the bound explicitly — pins and the active owner filter share the 50 slots, the unassigned bucket takes one, past ~50 combined a pin can still be dropped, and owner_truncated cannot distinguish that from tail truncation — plus what to render (a pinned-but-absent owner as zero).

limacharlie/commands/cloudsec.py
maximelb: Fixed in f286f84: confirmed against the classifier (a rule match sets Sensitive while Tier stays empty unless that rule assigns one), so 'no tier' really is a strict superset of 'not a crown jewel'. The help on all three flags now separates the two columns and --unclassified says so outright.

limacharlie/sdk/cloudsec.py
```diff
@@ -248,14 +331,94 @@ def get_finding_facets(
...
+        Returns:
+            ``{"causes": [{"key": str, "name": str, "kind": str,
+            "count": int}, ...], "distinct": int}``. ``kind`` is an OPEN
+            vocabulary (``firewall_rule``, ``entitled_identity``, ...) that
+            grows server-side — treat an unrecognized value as "some
+            object" rather than assuming a class. ``distinct`` is the total
+            number of causes matching the filter and MAY exceed
+            ``len(causes)`` when ``limit`` truncates.
+
+            A count is a whole-population figure, not a capped one: it does
```
maximelb, line 416: The rollup's POPULATION is the one caveat this otherwise obsessive docstring set omits, and it is the one a caller will get wrong.

`TenantStore.CauseFacets` reads all three of its queries against `cs_findings@{FORCE_INDEX=cs_findings_by_cause}` under `cause_key IS NOT NULL`, and a cause is stamped by exactly two kinds (`firewall_rule`, `entitled_identity`) on the exposure / attack-path / escalation detectors. go-cloudsec's own "cause kinds deliberately NOT minted" note records why the rest are out: the per-grant CIEM findings (1:1 with their finding), roles (not keyable), and the affected package on vulnerability findings — which that note puts at "~90% of an estate" — because the cause index is NULL_FILTERED specifically to hold only the cause-bearing attack-path slice. `PruneDegenerateCauses` then clears a cause that is unique and equal to its own finding's resource.

So a caller who reads "Takes the same filter selectors as `list_findings`, so a rollup can be scoped exactly like the list it summarizes" gets a wrong model: `list_finding_causes(finding_class=["vulnerability"])` returns `{"causes": [], "distinct": 0}` on an estate full of vulnerabilities — indistinguishable from "no such findings" — and under any filter the counts do not sum to what `list_findings` returns.

Suggest one clause here and in `_EXPLAIN_FINDING_CAUSES` (`limacharlie/commands/cloudsec.py:156`): the rollup covers only cause-bearing findings (today the exposure / attack-path / escalation slice); misconfig, vulnerability and per-grant CIEM findings carry no cause and are absent, so an empty rollup is not an empty worklist.

limacharlie/commands/cloudsec.py
```diff
@@ -1109,27 +1436,70 @@ def finding_list(ctx, severities, finding_classes, statuses, accounts,
 
 @finding_group.command("facets")
 @_finding_filter_options
+@click.option("--owner-pin", "owner_pins", multiple=True,
+              help="Keep these owners in the capped 'owner' facet even when "
```
maximelb, line 1462: The bounded-guarantee correction from `f286f84` reached only the SDK docstring. This `--help` still states the pin unconditionally, and so do `_EXPLAIN_FINDING_FACETS` (`:139`), `CHANGELOG.md` ("keeps named owners in the `owner` facet") and `doc/cli/cloud-security.md:80` ("keeps named owners in it even when they would not rank in").

The bound is exactly as reachable from the CLI as from the SDK: `facetOwner` builds `pinned = fs.Owners ++ fs.OwnerFacetPins` and truncates to `maxOwnerFacetValues` (50) — owners FIRST — and the gateway forwards up to `maxCloudSecFilterValues` (100) values per key, so `finding facets --owner ...` repeated 50+ times silently drops every `--owner-pin`, and `owner_truncated` cannot distinguish that from ordinary tail truncation.

This is the same one-surface-fixed divergence round 2 flagged for "the exact count". Suggest the same clause here and in the explain text / changelog / doc: the pin holds within the 50 slots it SHARES with the active `--owner` selection (the unassigned bucket takes one); past that render a pinned-but-absent owner as zero.

tests/unit/test_cli_cloudsec.py
```diff
@@ -1062,3 +1070,326 @@ def test_findings_empty_match_allowed(self):
...
+                return_value={"principals": []},
+            )
+            assert result.exit_code == 0, result.output
+            kwargs = inst.list_identity_access.call_args[1]
+            for key in ["admin", "external", "public", "disabled",
+                        "crown_jewel", "can_escalate", "dormant_90d",
+                        "with_sensitive", "mfa"]:
+                assert kwargs[key] is None, key
+
+    def test_identities_rejects_unknown_mfa_state(self):
```
maximelb: This is still the vacuous version round 2 called out: `f286f84` added the falsifiable sibling (`TestClosedVocabularyGuards.test_rejects_unknown_mfa_state_by_parse_failure`) NEXT to it instead of fixing it.

Verified by mutation on this branch: with `type=_MFA_CHOICES` (and the two tier/band Choices) deleted, the four `TestClosedVocabularyGuards` guards fail — and this test still passes, because unmocked it raises `AuthenticationError` and exits 1, which satisfies `!= 0`.

So the file now carries two tests for one behavior, one of which cannot fail while reading as coverage. Delete this one, or make it assert `exit_code == 2` and click's message like its sibling.

limacharlie/commands/cloudsec.py
```diff
@@ -898,6 +1069,144 @@ def _inventory_filter_options(f):
...
+             "— it is NOT 'off'.",
+    )(f)
+    f = click.option(
+        "--risk-band", "risk_bands", multiple=True, type=_RISK_BAND_CHOICES,
+        help="Filter by risk band; repeatable (OR). The band token the rail "
+             "renders, not a numeric range.",
+    )(f)
+    f = click.option(
+        "--unclassified", "unclassified", is_flag=True, default=False,
+        help="Include identities with no TIER assigned; combines with "
```
maximelb, line 1139: One more thing to say while this help is being precise about the two columns: the value `--unclassified` selects is the one selectable value with NO facet bucket, so the rail cannot report a count for a filter it can now apply.

- identity: `identityGroupFacet`'s drain skips `v == ""`, and the `criticality` histogram pass is additionally restricted to `IFNULL(i.is_sensitive, false)`.
- data-security: pass 3 of `ReadDataStoreFacets` is `WHERE criticality IS NOT NULL AND criticality != ''`.

With `--unclassified` active, `facets` narrows `total` correctly but its `criticality` map has no entry for the selection — the one exception to the claim this PR makes in `_EXPLAIN_CIEM_FACETS` / `get_identity_facets` / `doc/cli/cloud-security.md` ("a value's count is exactly how many rows selecting it would list"). Suggest a clause on this flag or in the facets explain text: the tier histogram reports only ASSIGNED tiers, so read `total` under the filter for the unclassified count — the same shape the `owner` facet already documents for its cap.

maximelb:
Round 3, re-reviewed against `f286f84`. All four round-2 findings independently verified as fixed: `--cause`'s `--help` and `CHANGELOG.md:32` now say "the count" (no "exact" left anywhere on that path); the guards are falsifiable — I deleted the three Choices plus `type=_MFA_CHOICES` and exactly the four `TestClosedVocabularyGuards` guards failed, restoring them turned them green; the `owner_pin` docstring now states the shared 50-slot bound and what to render past it; and the `--unclassified` / `--crown-jewel` / `--criticality` help now separates the tier column from the sensitivity flag, which matches the classifier (a rule match sets `Sensitive`, `Tier` stays `""` unless that rule assigns one). Also re-checked the rest of the diff: every documented CLI example parses (121 extracted invocations, no unknown option/invalid value), every `cs.*(...)` example in `doc/sdk/other-classes.md` matches a real signature and kwargs, no `--not-*` spelling survives, and no customer data, PII, private repo names or internal identifiers appear. Suite: 3810 passed, 5 skipped.

Open findings:
- `list_finding_causes` / `_EXPLAIN_FINDING_CAUSES` never say the rollup covers ONLY cause-bearing findings (`cause_key IS NOT NULL`, two stamped kinds) — vulnerability/misconfig/per-grant-CIEM findings are absent, so `--class vulnerability` returns an empty rollup on an estate full of them and counts never sum to the list.
- The `owner_pin` bound reached only the SDK docstring: `--help`, `_EXPLAIN_FINDING_FACETS`, `CHANGELOG.md` and `doc/cli/cloud-security.md` still promise it unconditionally, and 50+ `--owner` values drop every pin from the CLI too.
- `test_identities_rejects_unknown_mfa_state` is still the vacuous `exit_code != 0` version (passes with the Choice deleted); it now duplicates the falsifiable sibling added beside it.
- `--unclassified` selects the only value with no facet bucket (both criticality facet passes drop the empty tier), so the rail cannot report a count for a filter it can apply — the one exception to the "a value's count is exactly how many rows selecting it would list" claim.

limacharlie/sdk/cloudsec.py
maximelb: Fixed in 4560a3d: read the 'cause kinds deliberately NOT minted' note and documented the scope in both the docstring and the explain text — causes are stamped on the attack-path classes, vulnerability findings (~90% of an estate) carry none by design, so the counts never sum to the worklist total and --class vulnerability returning nothing means 'no shared fix on this class', not 'no findings'.

limacharlie/commands/cloudsec.py
maximelb: Fixed in 4560a3d: the bound is now on all four surfaces — CLI --help, _EXPLAIN_FINDING_FACETS, doc/cli/cloud-security.md and the changelog — not just the SDK docstring. Fair hit on the pattern: this is the second round where a correction landed on one surface out of four.

tests/unit/test_cli_cloudsec.py
maximelb: Fixed in 4560a3d: deleted it. It asserted only a non-zero exit, which an accepted value also produces without credentials, so it survived the mutation and duplicated the falsifiable sibling next to it; that sibling's comment now records why the weak form tested nothing.

limacharlie/commands/cloudsec.py
maximelb: Fixed in 4560a3d: documented on both --unclassified flags and in the SDK docstring — the empty-tier bucket is the ONE selection the facet rail does not count (both the identity criticality facet and the DSPM tier facet skip the empty string), so it is the exception to 'a facet count is exactly how many rows selecting it would list'.

lcbill (dismissed):

limacharlie/sdk/cloudsec.py
```diff
@@ -469,13 +760,90 @@ def get_topology(self) -> dict[str, Any]:
...
+
+        Args:
+            provider, account, region: Placement filters, OR'd within a
+                key. An empty-string value selects the unscoped bucket.
+            store_kind: Store kinds (``bucket``, ``sql_instance``, ...),
+                OR'd.
+            tier: Criticality tiers, OR'd. Same closed vocabulary as the
+                identity filter's ``criticality`` — ``critical`` /
+                ``high`` / ``medium`` / ``low``, plus the empty string for
+                stores with no tier assigned.
```
maximelb: The `--unclassified` facet blind spot reached the identity twin but not this one, so the SDK now documents the same selector two different ways.

Three lines up in `get_identity_facets`, `criticality` says the empty-tier bucket "is the ONE selection the facet rail does not count … the one case where a facet count cannot predict the list". Here `tier` says only "plus the empty string for stores with no tier assigned" — while this docstring opens with the strong claim ("so the counts **always** describe the population that list would return") and `list_data_stores` points readers *here* for what `tier` means. A DSPM SDK caller therefore never learns the exception.

The blind spot is identical on the store side and just as reachable: pass 3 of `ReadDataStoreFacets` is `WHERE criticality IS NOT NULL AND criticality != ''`, whereas the filter itself is `IFNULL(criticality, '') IN UNNEST(@tier)` — so `tier=[""]` narrows `Total` correctly and adds no `criticality` bucket, exactly like the identity `criticality` histogram (which drops `v == ""`).

Same surfaces still unqualified while making the claim:
- `get_data_security_facets` (this docstring) — the claim plus the `tier` arg
- `_EXPLAIN_CIEM_FACETS` (`limacharlie/commands/cloudsec.py:279`) — "a value's count is **exactly** how many rows selecting it would list", the strongest form of the claim anywhere in the PR, and the surface round 3 named
- `_EXPLAIN_DATA_SECURITY_FACETS` (`:367`) — "the counts always describe the population that list returns"
- `doc/cli/cloud-security.md:100` (the claim) and `:72` / `:87`, which introduce `--unclassified` on both groups with no caveat

This is the third round the same one-surface-fixed pattern lands ("the exact count" in round 2, `owner_pin` in round 3): one clause on each of those, or make the general claim say "except the empty-tier selection" once and reference it.

maximelb:
Round 4. Verified the four round-3 fixes against the backend, independently:

- Causes scope — correct. `CauseFacets`' three queries all read `cause_key IS NOT NULL` over the `NULL_FILTERED` cause index; only `firewall_rule` / `entitled_identity` are minted (`ClassToxicCombination`, `ClassPrivilegeEscalation`), and the "cause kinds deliberately NOT minted" note does put vulnerability findings at ~90% of an estate. Landed on both the SDK docstring and the explain text.
- `owner_pin` bound — correct and now on all four surfaces (`--help`, explain, doc, changelog). `facetOwner` concatenates `Owners ++ OwnerFacetPins`, clamps to 50 owners-first, and the unassigned bucket outranks every pin, so it costs a slot; a pin can be dropped from exactly 50 combined values.
- Vacuous `--mfa` guard — deleted; the surviving `TestClosedVocabularyGuards` sibling pins exit 2 + click's message.
- `--unclassified` facet caveat — correct where it landed (both CLI flags, `get_identity_facets`); see the one open item.

Also spot-checked, all accurate: empty `owner=` / `criticality=` / `tier=` survive the gateway (`len(v) > 0` presence gate, `addCloudSecMultiValue`'s one-element-array form), cause key clamp 512, rollup default 20 / cap 200, identity page default 500, store page cap 1000. `3809 passed, 5 skipped` reproduced (all 5 skips pre-existing/platform). No customer names, PII, private repo names or internal identifiers in the diff.

Open:
- `get_data_security_facets`' `tier` arg still documents the empty-tier selection without the facet blind spot its identity twin now carries — plus `_EXPLAIN_CIEM_FACETS`, `_EXPLAIN_DATA_SECURITY_FACETS` and `doc/cli/cloud-security.md:72/87/100`, which all still state the "a count predicts the list" claim unqualified. Third round of the same one-surface-fixed pattern. (inline)

limacharlie/sdk/cloudsec.py
maximelb: Fixed in 043ffae: swept for the claim rather than patching where it was spotted, since that is the third round this pattern cost us. The exception is now on every surface that made the unqualified claim — get_identity_facets, get_data_security_facets and its tier arg, the shared _identity_query_pairs comment, both explain texts (including _EXPLAIN_CIEM_IDENTITIES), _data_store_filter_options, doc/cli/cloud-security.md and the changelog — and `grep -rn 'always describe'` now returns nothing in the tree.

lcbill (approved):

maximelb:
Round 5 (final): no open findings. Round-4 fix verified against the backend — both facet passes really do skip the empty tier (`identityGroupFacet`'s `v != ""` drain, `ReadDataStoreFacets` pass 3's `criticality != ''`) while both filters are `IFNULL(...,'') IN UNNEST(...)`, and the exception now lands on all 12 surfaces with no unqualified variant left. Also checked the claim is *true*, not just consistent: the criticality histogram's extra `AND IFNULL(i.is_sensitive,false)` predicate does not add a second exception, because `classifier.classify()` returns `Classification{}` when `!Sensitive`, so a non-empty tier always implies sensitive. Routes/params/limits re-confirmed against lc_api-go and legion_graph (500 default page, 1000 store cap, 50-owner facet cap, 512-char cause clamp, 20/200 rollup, free-tier field set). 3809 passed / 5 skipped (all pre-existing platform+docstring skips). No PII or internal identifiers.

maximelb:
✅ AI-APPROVED

PR #360, opened 2026-09-08
feat(cloudsec): let a CI job state that its scanner ran, so a pushed SARIF can close

maximelb:
## The gap this closes

`go-cloudsec` v1.71.0 (merged today, from the AppSec code-lane pre-production review) moved a pushed SARIF's **closure authority** off the tool's rule *catalogue* and onto SARIF's own `run.invocations[].executionSuccessful`. That rule is correct: a scan step with a broken `--config` still emits the tool's full rule catalogue and zero results, and reading that as "the repository is clean now" resolves somebody's real vulnerabilities.

But it leaves a gap, and this repo is the natural place to close it.

**Trivy and Gitleaks do not write `invocations` at all.** A document straight from either is therefore *additive forever*: its findings land, nothing ever closes, and the customer's only symptom is a repository whose findings pile up with no explanation. This command's own documented example was `--source sarif -f trivy.sarif` — which, as of that release, can never close anything.

The missing fact is one the CI job **has** and nobody else does: whether the scan step exited 0.

## What this adds

`limacharlie cloudsec code ingest --scanner-succeeded / --scanner-failed` records that assertion in the document before pushing it. It is the *caller's* claim, attributable to the key that pushed it — not authority we inferred from a catalogue.

```
limacharlie cloudsec code ingest --repo acme/api --source sarif -f trivy.sarif --scanner-succeeded
```

Three properties worth preserving when this is next touched:

- **A run that already carries `invocations` is left alone, in either direction.** The tool's claim about its own execution outranks the caller's, and overwriting an `executionSuccessful: false` with `true` would manufacture exactly the authority the rule exists to withhold. The CLI says so on stderr rather than silently changing nothing.
- **`--scanner-failed` is a real value, not a no-op.** The flag is not a way to spell "always true".
- **A gzipped document goes back gzipped.** CI jobs compress large SARIF files, and handing the API a different encoding than it was given is its own bug.

Refused for `--source report` (which states its coverage directly) and `--source cyclonedx` (which makes no claim about a scan having run) — silently accepting it there would imply it did something.

Also updates the CLI docstring and the `ingest_code_results` SDK docstring, both of which now explain why a SARIF might close nothing and what `sarif_no_invocation_evidence` in `notes` means. That diagnosis was previously unavailable to anyone who hit it.

## Tests

Six new tests in `tests/unit/test_cli_cloudsec.py`:

| Test | Pins |
|---|---|
| `test_scanner_succeeded_records_the_evidence_that_lets_a_sarif_close` | the stamp lands, rest of the document untouched |
| `test_scanner_failed_records_the_failure_rather_than_omitting_it` | the flag is not "always true" |
| `test_a_tools_own_execution_claim_is_never_overwritten` | `executionSuccessful:false` survives `--scanner-succeeded`, and the CLI says so |
| `test_a_gzipped_sarif_is_stamped_and_stays_gzipped` | encoding round-trip |
| `test_the_flag_is_refused_for_sources_that_state_their_own_coverage` | non-SARIF refusal, nothing pushed |
| `test_without_the_flag_the_document_is_pushed_byte_for_byte` | the control — the CLI does not rewrite documents on its own |

**Evidence:** `pytest tests/unit/` → **4051 passed, 6 skipped, 0 failures**. (`tests/microbenchmarks/` needs `pytest-benchmark`, which is not installed on this machine; CI runs it.)

No new command, so `_COMMAND_MODULE_MAP`, `PROFILES` and `doc/cli/` need no change — this is an option on an existing verb.

### DIFF
--- a/limacharlie/commands/cloudsec.py
+++ b/limacharlie/commands/cloudsec.py
@@ -27,6 +27,7 @@ the one provider command here is the pre-save credential preflight
 
 from __future__ import annotations
 
+import gzip
 import json
 import os
 import subprocess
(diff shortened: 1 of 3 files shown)


lcbill (dismissed):

maximelb:
## Independent review findings — all six fixed in `f5292ce`

`/lc-review` does not run on this repository (it has never posted on a PR here), so I had the diff reviewed independently against `go-cloudsec@origin/master` and `legion_cloudsec_host@origin/master`. It found six real problems. The first made this PR's documentation actively misleading, which is worse than the gap it was written to close.

**1. The Gitleaks advice was false.** `--scanner-succeeded` cannot make a Gitleaks SARIF close anything, and no flag can. A gitleaks run is classified as a secret producer (`codescan/secretvocab.go`, `secretScannerDrivers`), every result falls through to `sarifClassSecret`, that branch only increments `droppedSecrets` and **never appends to the report**, and the class is then removed from the closure set outright (`codescan/sarif.go`, `delete(classes, model.CodeClassSecret)`). A pushed document cannot create a secret finding in any format, so there is nothing for a later push to close. An operator following the old text would add the flag, see `closed: 0` forever, and conclude the flag was broken. Both docs now say credential findings are never ingested (`secrets_not_ingestable`) and point at the hosted scan for that class. The Trivy half is genuine and confirmed against the checked-in real Trivy fixture (81 rules / 103 results, no `invocations`).

**2. The attribution claim was aspirational.** The docstring said the stamp was "attributable to the key that pushed it". Nothing implements that — the object is byte-identical to a tool-authored one, the server reads `invocations` as the *tool's* record, and stored provenance is the driver name. Claim removed and replaced with what is true.

**3. `"invocations": [{}]` was skipped.** A run with an invocation object that states nothing was treated as "the tool already spoke". The server reads a missing `executionSuccessful` as unsuccessful, so that document closes nothing — the CLI declined to stamp at exactly the moment stamping was needed. The test is now for the field, not the array.

**4. `ensure_ascii` inflation, against a cap applied after stamping.** The default escapes non-ASCII to `\uXXXX` — measured **1.94x** on a 1.1 MB CJK document — so a file comfortably under 20 MB on disk could be refused with a byte count matching nothing the operator can see. `ensure_ascii=False`, and when the stamp *is* what pushed it over, the error now says so rather than advising someone to narrow an already-small scan.

**5. The "nothing to do" message lied for an empty `runs` array** — it claimed every run already stated its result, which is reassurance, and the wrong one.

**6. A corrupt gzip escaped the error handling.** `BadGzipFile`/`EOFError` are not `ValueError`, so a truncated `.gz` printed a bare zlib message with no filename.

Also documents `sarif_execution_unsuccessful`, the note `--scanner-failed` produces, which neither doc named.

**Verified clean by the same review:** the stamped shape is exactly what `sarifRun.Invocations` / `runGrantsClosure` accept; absent `executionSuccessful` reads as failure (pointer-nil), so stamping is the only route; Click's tri-state with `default=None` behaves as assumed (proven by the byte-for-byte control, which would fail if the default were `False`); and the tests drive the real command and assert on the bytes handed to the SDK.

Four more tests added for the shapes those findings lived in. **4055 unit tests pass.**

lcbill (dismissed):

lcbill (approved):

PR #372, opened 2026-09-13
Add focused, evidence-backed D&R drafting workflows

maximelb:
Agents drafting D&R rules guessed dates, sensor IDs and array syntax despite available documentation. Add `dr prepare` to resolve a sensor, calculate a relative time window and package bounded real-event evidence with focused syntax guidance. Add `dr check` to explain structural mistakes and test each fixture scenario independently, distinguishing sample grounding and synthetic modifications from engine test results.

`dr deploy --workspace` rejects inputs that differ from the successful check, while retaining existing replay/metadata/etag verification. Nested fixture arrays support separate stateful scenarios. In agent mode, preparation activates a CLI drafting guardrail that allows stored evidence reads/checks but blocks endpoint tasking, scans and remote mutations until the next user request. This is explicitly not a sandbox for arbitrary Python/HTTP; independent isolation requires restricted credentials.

Validation: full SDK suite 4,284 passed / 5 skipped before one additional passing deployment regression test; 500 runner tests plus 8 subtests passed. New tests cover ambiguous identity, relative time, real array paths, malformed rules, fixture isolation, changed artifacts/evidence, stateful sequences and forbidden drafting operations. Staging-only live acceptance pending; no production deployment.


lcbill (dismissed):

maximelb:
Live restricted-key acceptance exposed LC_UID="" being sent as a user identity, causing organization API keys to fail with user-not-found. Normalized empty UID to absent and added regression coverage. Full suite: 4,286 passed / 5 skipped.

lcbill (approved):

PR #377, opened 2026-09-15
fix(hive): --expiry was unusable, because no value satisfied both ends

maximelb:
`limacharlie hive set --expiry` could not be used at all.

The flag validates Unix **seconds** (`validate_epoch_seconds`, and the CLI documents that convention in two other places), while the hive API stores and compares expiry in **milliseconds**. So a seconds value was accepted by the CLI and rejected by the API as already expired — `INVALID_EXPIRY_TIME` — and a milliseconds value was refused by the CLI before it was ever sent. There was no value a user could type that worked.

**The conversion belongs at the boundary that already claims a unit**, so the flag now multiplies by 1000 on its way out:

- every existing invocation keeps working;
- the guard stays — accepting both units would make the flag's own unit unknowable to a reader, since `1789459200` and `1789459200000` both mean 2026;
- zero still means "never expires" rather than the epoch, and needs no special case because it scales to itself.

**The flag only.** An `usr_mtd.expiry` supplied in an input file is passed through untouched: a record read back from the API already carries milliseconds, and converting it again would push the expiry some 55,000 years out. `hive get ... | hive set ...` has to round-trip.

**Two help topics corrected**, because they are part of why this went unnoticed: the hive record description gave no unit for `expiry`, and the timestamps topic asserted that *all* API timestamps are seconds — exactly the assumption that makes this flag look fine.

**Tests:** four, each pinning one decision — the conversion happens, zero stays zero, a milliseconds value is still refused, and a file-supplied expiry is not converted. Reverting any one of them turns a test red (verified).

Full unit suite (4144 passed, 19 skipped) and microbenchmarks pass.

Found while building the Email Security acceptance driver, where every live run died on `INVALID_EXPIRY_TIME` before sending anything.


lcbill (dismissed):

maximelb:
Reworked on an independent review; head is `ac1a1d4`. It found that the first version fixed the flag and left the documentation people actually read still saying the old thing — and that one claim in my commit message was false.

**Accepted and fixed**

- **The flag's own `--help` had become literally false** ("sets usr_mtd.expiry to this value" — it no longer does), and every `--ai-help` block still printed `expiry: 0  # unix epoch` with no unit. That is what an agent driving this CLI consumes. Corrected in the flag help, in `_EXPLAIN_LIST`/`_EXPLAIN_GET`/`_EXPLAIN_SET`, and in `fp.py`, whose prose recommends expiry for temporary FP rules and never said which unit.
- **The new timestamps topic was wrong in a new way.** I called hive expiry "the documented exception" to the seconds convention; `audit list`'s `time`, `event overview`'s timestamps and search's `mtd.ts` are milliseconds too. It now lists them.
- **A negative expiry was accepted and amplified** — `-1` went out as `-1000` and came back as the same opaque `INVALID_EXPIRY_TIME` this flag exists to stop producing. Guarding only the upper direction was the asymmetry.
- **The conversion was silent.** `hive get` renders the stored value, so a user typed `1789459200`, read back `1789459200000`, and had nothing to explain it. `hive set` now says what it sent.
- `doc/cli/hive-data.md` gains an Expiry section, and loses four examples of a `--category` flag that does not exist (pre-existing; fixed while in the file).

**My claim was false, and the review proved it empirically.** I wrote "reverting any of the four turns one red"; reverting the fix turns exactly ONE red. `test_zero_still_means_never` could not fail against any implementation — 0×1000 is 0, and there is no zero-handling code to revert — so it is folded into a real test rather than standing as a tautology. The override branch (flag AND input file) had no test at all; it has one now, with two values deliberately not a factor of 1000 apart, since round numbers would let it pass whether the flag won or the file did.

**Noted, filed rather than fixed here:** `--expiry` exists only on `hive set`, so the shortcut commands (`fp`, `lookup`, `dr`, `secret`, …) cannot set one — which is awkward given `fp`'s help is where expiry is recommended. And `hive validate` cannot dry-run it. Both are feature gaps rather than this bug, and widening the flag's surface belongs in its own change.

Six tests, four mutants, each red. 4146 unit tests and 90 microbenchmarks pass.

/lc-review

lcbill (approved):

PR #379, opened 2026-09-17
mailsec: --force to perform a remediation in alert-only mode

maximelb:
## Summary

An Email Security org with no automation in enforce mode is in **alert-only** mode: every remediation is decided and audited but withheld, and the API returns `result: alert_only`. Until now, people acting from the CLI or SDK were withheld the same way and had no way to override it. The backend now accepts an explicit per-action `force` boolean. This PR exposes it.

### SDK (`limacharlie.sdk.mailsec.Mailsec`)
- `act_on_message(..., force=False)` → `POST .../messages/{msg_uuid}/actions`
- `act_on_campaign(..., force=False)` → `POST .../campaigns/{campaign_id}/actions` (execute only)
- `bulk_action_execute(..., force=False)` → `POST .../actions/bulk/execute`

The body gets `"force": true` only when `force` is `True`. If it is absent or false, the request body is byte-identical to before. `force` is not part of any confirmation token. Passing `force=True` to `act_on_campaign` without a `confirm` (absent or empty, i.e. a preview) raises `ValueError` rather than dropping the flag silently. A non-bool `force` raises `TypeError`.

### CLI
- `limacharlie mailsec message action ... --force`
- `limacharlie mailsec message bulk-action ... --confirm <token> --force`
- `limacharlie mailsec campaign action ... --confirm <token> --force`

Flag help: "Perform the action even if the organization is in alert-only mode (no automation in enforce mode). The override is recorded in the audit trail."

On a preview (no `--confirm`), `--force` is not sent, and the command prints a note saying it applies to the execute. `--reason` already works this way.

When a single action or campaign execute response, or a settled bulk job's status, has `force_required: true`, a note goes to stderr (suppressed by `--quiet`):

> This organization is in alert-only mode, so the action was recorded but not performed. Re-run with --force to perform it.

For bulk, the note adds that the same `--confirm` token works and that a forced execute runs as a new job. `mailsec message bulk-status` prints the same note. If `--force` was sent but the response still has `force_required: true` (an API that predates `force` drops the field), the note says force was not applied instead of telling the caller to use `--force` again. Stdout still carries the server's document unchanged, so the JSON shape is the same.

`--ai-help` text for the three commands and `doc/cli/email-security.md` describe the flag.

## Test plan
- [x] SDK wire tests: `force` is present only when `True` (message, campaign execute, bulk execute); a forced campaign preview is refused before any request
- [x] SDK: a non-bool `force` and a forced campaign preview with `confirm=""` are refused before any request
- [x] CLI tests: `--force --no-wait`; force sent but not applied (message, campaign, bulk-action, bulk-status); flag forwarded; not forced by default; not sent on a campaign or bulk preview, with a note; `force_required` note printed (message, campaign, bulk) and not printed otherwise; stdout document unchanged; `--quiet` silences the note; help wording
- [x] Mutation check: each new assertion was confirmed to fail against a reverted/mutated implementation (34 mutants, all killed)
- [x] `pytest tests/unit/ tests/microbenchmarks/ --benchmark-disable`: 4297 passed, 5 skipped (pre-existing platform skips)


lcbill (dismissed):

maximelb:
## Adversarial self-review (pristine `git archive` extraction)

This repo does not use the `/lc-review` bot, so the diff was reviewed from a clean extraction of `09ff42e`. The review found no blockers. It checked the field names against the backend (`force` on the request; `force_required` on single-action and campaign-execute responses; `force` and `force_required` on bulk status), whether force could leak onto a preview, whether stdout changed, request bodies for existing callers, exit codes, and `--quiet`.

Findings and what was done (fixed in `bebc41e`):

1. **should-fix (fixed)**: when `--force` was sent but the response still had `force_required: true`, the note said "Re-run with --force". That happens when the API in use predates `force` and drops the field. For bulk it also contradicts the "adopting existing job" line. The note now says `--force` was sent but not applied. Bulk also checks the job's own `force` field. Tests cover message, campaign, bulk-action and bulk-status.
2. **should-fix (fixed)**: `act_on_campaign(confirm="", force=True)` was not refused, but an empty `confirm` is a preview on the server. The guard now uses `not confirm`, with a test.
3. **nit (fixed)**: a non-bool `force` (`1`, `"true"`) did not force and gave no error. All three SDK calls now raise `TypeError` before any request, with a test.
4. **nit (fixed)**: `mailsec message bulk-status` now prints the same `force_required` note.
5. **nit (fixed)**: the `campaign action` `--ai-help` examples now go preview, then execute, then forced execute.
6. **nit (fixed)**: the `--quiet` test now has a positive control. Added a test that `--force --no-wait` still forwards force.
7. **not changed, existing behaviour**: a bulk job whose members were all withheld (`ok=0, failed=0, alert_only>0`) settles with exit code 0. The new stderr note tells a person at a terminal, but a `&&`-chained script cannot see it from the exit code. Changing the exit code is outside this PR.

Deployment note: the `--force` flag only has an effect once the API gateway forwards `force` on the three action routes.

Verification: `pytest tests/unit/ tests/microbenchmarks/ --benchmark-disable` gives 4297 passed, 5 skipped (platform skips that were already there). Mutation check: 34 mutants against the new behaviour (SDK wire, CLI forwarding, preview exclusion, notes, quiet, help wording, bool validation, not-applied message), and every one was killed by its targeted test.

lcbill (approved):

PR #380, opened 2026-09-17
cloudsec code scan: run static-analysis rules with scanner v0.16.0

maximelb:
## What was asked

Keep `limacharlie cloudsec code scan` running static-analysis (`sast`) rules with the new code scanner. From v0.16.0 the scanner runs only the rule set it is given: `--default-rules` (LimaCharlie's default set) or `--rules-file` (a rule set document). Older scanners reject both flags.

## What was done

- The pinned default scanner image moves from v0.4.0 to v0.16.0.
- When `sast` is in `--scanners`, the CLI passes `--default-rules`. Scans without `sast` pass no rule-set flag, so they still work on older scanners.
- New `--rules-file PATH` runs a rule set document (`{"version":1,"records":[{"key":...,"rules":{"rules":[...]}}]}`) in place of the default rules. Before the scan starts, the CLI refuses a document the scanner would not accept (invalid JSON, unknown fields, a version other than integer 1, a record without a unique non-empty key or a `rules` object, over 32 MiB) and a document with no rules. The scanner exits 0 even when it refuses a rule set, so without this check a bad file would look like a clean scan. The file is copied into the scan's private scratch directory, so no directory you name gets mounted into the container.
- New `--org-rules` reads the org's `cloudsec_code_rule` hive records and builds the rule set a hosted scan of that org runs: enabled, unexpired records only, sorted by key, each contributing just its `rules` array, written in the same field order the hosted service uses. The CLI warns about enabled records that have no rules, refuses an org that has no usable rule before the scan starts, and reports hive read errors clearly.
- `--rules-file` and `--org-rules` cannot be combined, and both are refused when `sast` is not requested.
- A custom `--image`/`--binary` that exits with a usage error (exit 2) while a rule-set flag was passed now gets an error message saying that `sast` needs scanner v0.16.0 or newer.
- Added a "Code scanning: local scans" section to `doc/cli/cloud-security.md`, and updated the command help and `CHANGELOG.md`.

## Testing

- `pytest tests/unit/ tests/microbenchmarks/`: 4280 passed. The 5 skips are existing platform-specific tests (Windows/macOS only, and two documented docstring gaps). This PR adds none.
- 14 new unit tests mock the subprocess/docker boundary and check the arguments built in each mode: default rules (container and binary), no `sast`, `--rules-file` (container path and byte-for-byte content, binary host path), `--org-rules` document composition (enabled/disabled/expired/empty records, sort order, only the `rules` array), an org with no rules, no hive read when a local check fails, a hive read failure, invalid rule files, flag conflicts, and the old-scanner usage hint (shown only on exit 2).
- Ran against the real images locally on a small Python tree:
  - v0.16.0 with default rules: exit 0, report shows `sast_rules_count` 176 and 2 findings.
  - v0.16.0 with `--rules-file` holding one custom rule: exit 0, `sast_rules_count` 1, and the finding comes from that rule.
  - `--image` pointing at v0.4.0 with `sast`: the scanner prints its unknown-flag usage message and the CLI error names v0.16.0 as the minimum.
  - An `--org-rules` document built from mocked hive records (one enabled, one disabled): the scanner loaded 1 rule and reported a finding from it.
- `--org-rules` was not run against a live org's hive, only against mocked records.
- Adversarial self-review findings and how each was resolved are posted as a PR comment.


lcbill (dismissed):

maximelb:
Self-review (adversarial read of a pristine extraction of 2f9bf9c). This repo does not use the AI review bot.

Confirmed correct: container vs binary argv and rules-file placement (`<workdir>/rules.json` is `/scan/rules.json` in the container), no rule flag without sast, conflicting flags refused before any work, the hive read happens after local checks, the Hive SDK usage and filtering, the exit-2 hint limited to a custom `--image`/`--binary`, and the tests go through the real exit handling.

Findings and resolution (fixed in the follow-up commit):
1. **Medium, fixed.** `--rules-file` checks were looser than the scanner's parser, which refuses unknown fields. The scanner reports a refused document as a section error but still exits 0, so the scan looked successful. The CLI now refuses unknown top-level and record fields, a non-integer `version`, and documents over the scanner's 32 MiB limit.
2. **Medium, fixed.** A rules file with no rules was accepted, which `--org-rules` already refuses. Both now refuse it.
3. **Low, kept.** Exit 2 also covers scanner configuration errors, so the old-scanner hint can show up for those. The hint is worded as conditional ("if the scanner reported an unknown flag"), and the scanner's own message is printed above it.
4. **Low, partly fixed.** The `--org-rules` document is now written in the hosted service's field order (version, then records; key, then rules; sorted rule keys; no HTML escaping). The docs still only promise the same rules, not a byte-identical digest.
5. **Low, fixed.** Added a test that `--org-rules` makes no hive read when a local check fails.
6. **Info, intended.** `--scanners sast` against a custom scanner older than v0.16.0 now fails with a clear message. Those scanners do not accept the rule-set flags, and v0.16.0+ runs no sast without them.

Additionally ran end to end against the real v0.16.0 image with an `--org-rules` document (hive mocked): 1 enabled record loaded, the disabled one excluded, and the finding came from the org rule.

lcbill (approved):

PR #382, opened 2026-09-18
cloudsec: bind the Code Security GA gateway surface

maximelb:
An audit of `limacharlie cloudsec` against the public API after the Code Security GA turned up six live route groups the CLI never reached, and four findings selectors it dropped. This binds all of them, with tests and docs.

## What was missing

**Container images had no binding at all.** Four live routes (`/code/image-repos`, `/code/image-repos/facets`, `/code/images`, `/code/images/{digest}`) — a CLI user could not see a single container image.

**Two code-lane write doors were unbound.** `POST /code/pr_check` is the documented developer door a CI job is meant to call; `POST /code/webhook` is the "Fix webhook" door, and without it push rescans and pull-request checks cannot be wired up from the CLI at all.

**The compliance v2 family was unbound** — immutable runs, attestations, the drift stream, deterministic exports and schedules. The CLI only had the live, point-in-time report.

**`GET /azure/scope-hierarchy` was unbound.**

**Four findings selectors were dropped**, and one of them matters more than its number suggests. `grain` is the only worklist selector whose ABSENCE is not "unconstrained": it re-engages the server's leading-grain default, which excludes the per-CVE rollups a package finding already states pair for pair. So the CLI **could not reach per-CVE findings at all**, and had no way to pivot from a container image to its findings. `fix_state`, `exploit_band`, `image_urn` and `cause` ride the same four routes (`finding list` / `facets` / `causes` / `export findings`).

## What this adds

| | |
|---|---|
| `cloudsec image repos\|repo-facets\|list\|get` | the registry-backed container-image inventory |
| `cloudsec code pr-check` | what does this pull request introduce, as a check run |
| `cloudsec code webhook` | point a connection's GitHub App webhook at the org's adapter |
| `cloudsec compliance run\|runs\|attestations\|attest\|events\|export\|schedules\|schedule-set` | audit-grade compliance |
| `cloudsec azure scope-hierarchy` | Azure scope containment evidence |
| `--grain` `--fix-state` `--exploit-band` `--image-urn` `--cause` | on `finding list`, `finding facets`, `finding causes` and `export findings` |

## Decisions worth a second look

- **Closed vocabularies are validated client-side** (`--grain`, `--fix-state`, `--exploit-band`, and the image `--sort` keys) rather than forwarded. The failure modes differ: a bad repeatable value returns an honest empty page, but a dropped `--grain` value re-engages the default and answers a *different* question successfully. The image sort keys are validated because the server silently coerces an unrecognised key to `name`, which would hand you a successful, wrongly-ordered page.
- **`--action edited` requires `--prev-base-sha`, refused client-side.** `edited` also reports a title or body change, which changes nothing about what a PR introduces. Refusing it here means a repository whose descriptions are being edited does not turn into published RPCs.
- **Tri-state flags stay tri-state.** `--with-findings`/`--with-images`/`--running`/`--signed` default to `None`, never `False` — the negative is a real selection on all four. `--signed` additionally cannot see images whose signing status was never reported, and the help says so.
- **`compliance export -o` decodes the base64 `content`** before writing, so the file holds the document rather than its encoding.
- The digest in `image get` is percent-encoded; verified against the gateway, which matches on the escaped path and unescapes path params.

## Discovery + docs rot fixed

`cloudsec code capabilities` and `cloudsec code fixes` shipped in #378 without ever reaching a discovery profile, so `limacharlie help discover` could not surface them — invisible to anyone, or any agent, finding the CLI that way. They are listed now, and this adds the **two-way coverage test `mailsec` already had**, so the cloudsec surface and the discovery map cannot drift apart again in either direction.

The CLI reference documented only the *local* `code scan`; the entire hosted code lane was undocumented. It now covers the lane, pull-request checks, webhook wiring, container images, the vulnerability selectors and compliance v2.

## Testing

`pytest tests/unit/ tests/microbenchmarks/` — **4367 passed, 6 skipped**, up 50 tests. The new tests pin the wire shape of every new route, the tri-state defaults, the cursor walks, the client-side refusals, and that the webhook secret is never echoed.


lcbill (dismissed):

lcbill (dismissed):

maximelb:
Addressed the review findings; pushed as f42ffc5.

**Two real contract bugs, both confirmed against the server before fixing:**

1. **`--action` was optional on `code pr-check`, and the host refuses a body without one.** The gateway guards its vocabulary check with `action != ""`, so an absent action passes it — but the collection host tests membership of the closed set with no such exemption, and `""` is not a member. So the request was refused unconditionally, and the shape that failed is precisely the one a CI job would send: the caller this command exists for. Now required, with `synchronize` named in the help as what a push to an open pull request is.

2. **`--scanning-state` was repeatable, but only the first value survives.** The backend takes a list; the gateway forwards that key through its scalar loop, so a second filter value vanished with no error — a confidently wrong page. Made a scalar with the reason in the help. The gateway is where this should really be fixed; until it is, the CLI doesn't claim a filter it can't deliver.

**Docstring/help claims corrected where they didn't match the server:** `rationale` and `approved_at` aren't required on an attestation (and one written without `approved_at` is stored but never counts toward a control — worth more than a "required" label); `low_coverage` is only computed when `applicable` is true; `compliance events` falls back to its *default* rather than clamping when `days`/`limit` exceed the ceiling; a superseding attestation revision need only be higher, not `+1`, same for a schedule revision; the webhook url's adapter segment is enforced by prefix, not by the full connection name; and every repeatable selector is silently truncated at 100 values.

**Also:** `compliance export -o` now writes bytes it's handed directly instead of failing on them, so a transport that doesn't base64-encode still produces the file.

One test had pinned a pr-check body with no action as the supported minimal form — the request the host always refuses. That's what let the contract bug through; it now asserts the opposite.

`pytest tests/unit/ tests/microbenchmarks/` — 4369 passed, 6 skipped.

lcbill (approved):

